PR #10, opened 2018-07-17
trig rules lua serialization syntax fixed

shdwp:
This uses lua serializer dumps to correctly output arguments for trig rules inline lua code. 

Primary error was that string arguments for conditions were serialized without quotes, but I find some another places that could potentially bring the same issue.

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -1,3 +1,4 @@
+from .lua.serialize import dumps
 from .translation import String
 from enum import Enum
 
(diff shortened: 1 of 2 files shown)


dcs/action.py
```diff
@@ -11,7 +12,7 @@ def __repr__(self):
         s = []
         for x in self.params:
             if isinstance(x, String):
-                s.append('getValueDictByKey("' + x.id + '")')
+                s.append("getValueDictByKey({})".format(dumps(x.id)))
             else:
                 s.append(str(x))
```
shdwp: Could you elaborate - what type `parameters` would be? I didn't find out the type for it, I assume it would be an condition instance?

rp-: you mean the self.params?
that can be pretty any type, for example for the MessageToCoalition it is
coalition identifier: the coalition (red, blue, ...)
String() to display
number of seconds how long the message is displayed
bool if the text view should be cleared

shdwp: I mean, than the str(x) should be changed to dumps too!

rp-: you maybe right, but I think here it is not a big problem because the string that will be added here, are pretty static

PR #11, opened 2018-07-20
ICLS + TACAN update

shdwp:
Subject. 

I haven't found any docs on "unitId" in those tasks, but it seems that without it they would simply not work.

Stennis seems to use system 3 rather than 4, so I've added the aa argument to support both.

### DIFF
--- a/dcs/task.py
+++ b/dcs/task.py
@@ -859,19 +859,37 @@ class ActivateBeaconCommand(WrappedAction):
 
         return freq * 1000000
 
-    def __init__(self, channel=10, modechannel="X", callsign="TKR", bearing=True):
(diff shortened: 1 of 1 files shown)


shdwp:
Also it could be that that `aa` option should be default or something like this, since it looks like tankers will require it too (default tacan that gets generated by refuel flight is not working too).

Do you have any comments on this?

rp-:
never seen this aa option, maybe it's new.
As you noticed there isn't really any documentation about anything, it is a lot of guess work and trail and error from the mission editor.
the best documentation you can find is from MIST
https://wiki.hoggitworld.com/view/Mission_Scripting_Tools_Documentation
because they basically only add the structures from mission file dynamically.
But that doesn't cover all parts of a mission file.

PR #12, opened 2018-07-29
Remove previous loadout before loading new one

shdwp:
I don't know whether this is by design, but it does seem counter intuitive anyway.

+ that dumps update


rp-:
Thanks

PR #19, opened 2019-06-27
Fix error in unittype.py for DCS World Steam Edition users

Khopa:
Hello,

I was having an error in 'unittype.py' when using "dcs_liberation" latest version (which is using pydcs). 

The error occured when the code was trying to determine the DCS install directory through the registry here :

https://github.com/pydcs/dcs/blob/5c33c556880247068e7c452ce539670bde976e8c/dcs/unittype.py#L65-L72

The thing is that for Steam edition users, there does not seem to be a "Path" value defined for "HKEY_CURRENT_USER\Software\Eagle Dynamics" in the registry to locate the DCS World installation folder.  

image

I could have defined the registry value manually, but then every Steam users using pydcs would have had to do the same which is not ideal.

So i wrote a little module to abstract the search of the install directory (whether it's a steam or standalone version) and retrieve various infomations about the local DCS World installation. For standalone version users, i'm using the same code as before; and for steam users, it's a bit more difficult to find the install directory. The method is to : 

* Find the Steam install directory using the windows registry
* Find the directories where Steam can install games which is defined in a file in Steam installation directory
* Search for a DCSWorld folder in these directories.

(See the new installation.py file)

**Disclaimer :** 
I haven't tested this with DCS Standalone Edition installed on my computer, as i don't have it and didn't really wanted to risk ruining my own 100+Gb Steam DCS installation by installing it; 😁

### DIFF
--- a/dcs/__init__.py
+++ b/dcs/__init__.py
@@ -18,5 +18,6 @@ from . import triggers
 from . import condition
 from . import action
 from . import forcedoptions
+from . import installation
 from .mapping import Point, Rectangle, Polygon
 from .mission import Mission
--- a/dcs/installation.py
+++ b/dcs/installation.py
@@ -5,8 +5,14 @@ TODO : add method 'is_using_open_beta', 'is_using_stable'
 TODO : [NICE to have] add method to check list of installed DCS modules (could be done either through windows registry, or through filesystem analysis ?)
 """
 
-import winreg
 import os
+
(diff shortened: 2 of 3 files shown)


rp-:
I'll check the PR on the weekend, currently on vacation ;)

rp-:
I squashed/fixupped your branch into 1 commit and pushed it to master
Thanks!

PR #23, opened 2019-10-14
fix(dcs check): correctly detect openbeta DCS installs

wrycu:
Current version fails to find any DCS version if the user has installed the openbeta as their primary install.  This request looks for the beta version in addition to the normal.


rp-:
fixed and merged

PR #27, opened 2020-03-06
WIP: .miz loading improvements; updated lua parser for F14 loadouts

shdwp:
### Lots of minor additions to support some specific `.miz` files. 
I've loaded mission file of *GAW*/*PGAW* and those were the only fixes I needed to implement. A couple of new options as well, but I was too lazy to add enums for them as well.
Good job @rp- , those were the only fixes I needed to implement, and mission file was 112k lines long!

### Updated `lua` parser to support syntax in `F-14B` UnitPayloads
It now includes globals dict to pass during parsing, supports local variable assignments to for down the line values substitution, multiple assignments syntax and `return` statements.

### Preserve miscellaneous files in archive
So pretty much anything that was in loaded `.miz` will be included during saving. The missions that I've loaded included stuff like `Config` and `Kneeboard` folders which were ignored during loading.

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -1432,7 +1432,8 @@ class AITaskPush(Action):
 
     @classmethod
     def create_from_dict(cls, d, mission):
-        return cls(d["ai_task"][0], d["ai_task"][1])
+        values = list(d["ai_task"].values())
+        return cls(values[0], values[1])
 
     def dict(self):
         d = super(AITaskPush, self).dict()
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -492,6 +492,30 @@ class Explosion(Action):
         return d
 
 
+class Smoke(Action):
+    predicate = "a_effect_smoke"
+
+    def __init__(self, zone="", density=1, preset=1):
+        super(Smoke, self).__init__(Smoke.predicate)
+        self.zone = zone
+        self.params.append(self.zone)
+        self.density = density
+        self.params.append(self.density)
+        self.preset = preset
+        self.params.append(self.preset)
+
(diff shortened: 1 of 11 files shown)


shdwp:
One minor thing - airframe export leaves out `P_51D_30_NA` without any liveries included, leading to runtime error. I've simply added the class, but it probably needs to be fixed in the exporter.

Also apparently lua tests on Travis fail, I may have missed something, I'll rerun them and see what exactly fails.

shdwp:
For now it's still `WIP` since I would also like to add support for another few missions.

rp-:
puh, thanks for the fixes, but it would be easier if those commit would have been split into multiple pull requests. And maybe a info where I could get the GAW mission, so I could verify the stuff on my end too.
This way I can't easily hit the "rebase and merge" button, because I will not commit something that is "WIP". I don't say it must be perfect, just that the commits are logic and structured.

shdwp:
Sure, that was the whole point of puttin `WIP` in the title of the PR. Once it's ready I'll give you a green light. 

Here's the repos of the missions I've been testing it against: 
* https://gitlab.com/hoggit/developers/persian-gulf-at-war
* https://gitlab.com/hoggit/developers/georgia-at-war
* https://gitlab.com/hoggit/developers/georgia-at-war-phase-2

rp-:
Oh, sorry than I misunderstood that

rp-:
I've cherry-picked the lua parser stuff and added a bit on it and pushed that to master FYI

shdwp:
New PR: https://github.com/pydcs/dcs/pull/28

PR #28, opened 2020-04-02
Loading improvements

shdwp:
Same PR as before, but I've omitted parser commit and rebased to current master.

One thing I'm not sure about is that I've changed constructor signature of `TriggerStart` and `TriggerCondition` to be in line of the other `Trigger` classes, and I'm not sure whether that was the mistake or it should've been fixed elsewhere (`TriggerRule/create_from_dict` will always invoke constructor with two arguments - `event` and `comment`). I don't have enough time to investigate so it would be great if you can comment on this.

### Lots of minor additions to support some specific .miz files.
I've loaded mission file of GAW/PGAW and those were the only fixes I needed to implement. A couple of new options as well, but I was too lazy to add enums for them as well.
Good job @rp- , those were the only fixes I needed to implement, and mission file was 112k lines long!

### Preserve miscellaneous files in archive
So pretty much anything that was in loaded .miz will be included during saving. The missions that I've loaded included stuff like Config and Kneeboard folders which were ignored during loading.

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -1432,7 +1432,8 @@ class AITaskPush(Action):
 
     @classmethod
     def create_from_dict(cls, d, mission):
-        return cls(d["ai_task"][0], d["ai_task"][1])
+        values = list(d["ai_task"].values())
+        return cls(values[0], values[1])
 
     def dict(self):
         d = super(AITaskPush, self).dict()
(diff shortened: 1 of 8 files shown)


rp-:
Not sure why I made it different with TriggerStart, maybe I was sure there is never an Event. but the change shouldn't harm that.

I just made a few cleanups in the commits (merge conflict was committed(first commit)) and rebased it to master

shdwp:
Weird, I remember fixing that and pushing the branch. I guess it doesn't matter now, thanks.

PR #35, opened 2020-06-07
Added support for trigger actions : MarkToAll, MarkToCoalition, MarkT…

Khopa:
Hello, 

I added support for these trigger actions : 

* MarkToAll
* MarkToCoalition
* MarkToGroup 

image

These allow to create "marks" that will appear in DCS on the F10 map :

image

See the new test in test_mission.py for an example usage. 

Khopa

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -669,6 +669,124 @@ class LoadMission(Action):
         return d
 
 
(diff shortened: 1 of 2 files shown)


rp-:
Thanks again!

PR #37, opened 2020-06-08
Added support for TheChannel map + data export for P-47

Khopa:
* Added The Channel map
* New data export for openbeta 2.5.6.50321  (includes the new P-47D-30 module)
* Fix : The FW-190A8 was not marked as a flyable module


rp-:
:clap: :+1:

Issue #50, opened 2020-07-16
Various airports missing from Normandy


wrycu:
Attempting to parse a Normandy mission results in `AttributeError: 'NoneType' object has no attribute 'load_from_dict'`.

After looking through the code and map, the following airports are missing:
- Conches
- Argentan
- Goulet
- Vrigny
- Essay
- Hauterive
- Barville

You can reproduce by saving any mission on the Normandy map, but just in case, here's one anyway.
[normandy.zip](https://github.com/pydcs/dcs/files/4933457/normandy.zip)

Khopa:
I made a new airport dump for Normandy. 
See https://github.com/pydcs/dcs/pull/52

But i'm still worried about possible compatibility issues since they changed one of the airport name in game. Hopefully it won't cause any issues.

PR #52, opened 2020-07-17
Added Normandy missing airbases + data-export 2.5.6.52196

Khopa:
The airport dump for Normandy had not been made since the last map update in late 2019 iirc.
See https://github.com/pydcs/dcs/issues/50

I ran a new airport dump and added the missing airports.

I noticed :
- a lot of changes to runway directions, parking slots positions... etc
- The name of the Ford base has changed. (Now "Ford AF" instead of "Ford")

***

I also jumped on the occasion to make a new data-export to add the few units that were added in the very last open beta.

By analyzing the new export, i noticed that:
- Deka added a new Chinese ship  (Type 071)
- Deka changed the id for almost all their ships. (I'm worried about compatibility issues.)
- There are more possible payloads for the JF-17 
- ED added a new MLRS unit

### DIFF
--- a/dcs/countries.py
+++ b/dcs/countries.py
@@ -22,6 +22,7 @@ class Russia(Country):
             SPH_2S19_Msta = vehicles.Artillery.SPH_2S19_Msta
             MLRS_9A52_Smerch = vehicles.Artillery.MLRS_9A52_Smerch
             MLRS_9K57_Uragan_BM_27 = vehicles.Artillery.MLRS_9K57_Uragan_BM_27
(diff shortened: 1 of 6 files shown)


rp-:
Thanks!
but before I merge this, we should check on that changed id's what DCS is really using in the .miz files.
Maybe they still use the old ids and the export script needs an update and otherwise we may have to make the id matching not too strict while loading the .miz file.

rp-:
Ok I checked it now.
DCS editor uses the new Type_* id, and if you load an old mission without "Type_" it ignores the unit at all.
So it is save how it is now.

Thanks

PR #64, opened 2020-08-10
Add manual heading

Khopa:
Hello @rp- 

I made this branch to add the "manualHeading" parameter. This is this parameter in DCS mission editor, that is only available for ground units : 

image

This allow forcing the heading of a unit regardless of its waypoints.

I also made some small changes to some "Option" tasks, and completed some Enums.

Thanks,

Khopa


rp-:
Cool PR, Thank you!

PR #70, opened 2020-09-01
Add support for setting preset radio channels.

DanAlbert:
One behavioral change here that is present even without using the new
API: channel 1 of radio 1 will no longer be forcibly set during
serialization. It will still be set when set_frequency() is called,
but this allows callers to set their own channel 1 for radio 1 as
desired. DCS (the mission editor, at least) will automatically set
channel 1 of the first compatible radio to the flight's frequency, but
that is not necessarily the first radio (such as for an F-16 with a
VHF flight channel).


rp-:
Thanks

PR #73, opened 2020-09-18
Reduce differences between DCS and pydcs mission files.

DanAlbert:
DCS Liberation currently has an issue where generated missions need to be opened in the mission editor and saved before delayed flights behave properly. I'm not completely sure that either of these changes fixes that issue (I can't reproduce it myself), but these seem to be the only two behavioral differences in the mission file before and after saving with the mission editor. They do seem to be incorrect, although it's possible these are working due to some default behavior for each function.


rp-:
Thanks

PR #75, opened 2020-09-24
Dataexport

Khopa:
Hello, 

I made a data-export with the latest version. 
This is mainly to support new loadout options on the F-16C.
This also add new P-47 variants, and the CVN-75 supercarrier.
Lebanon, and a bluefor and redfor countries were also added.

Also, not sure why, but it looks like most livery ids have been changed to lowercase in game, not sure what is going on.

Khopa

### DIFF
--- a/dcs/countries.py
+++ b/dcs/countries.py
@@ -177,6 +177,9 @@ class Russia(Country):
         SpitfireLFMkIXCW = planes.SpitfireLFMkIXCW
         P_51D_30_NA = planes.P_51D_30_NA
         P_47D_30 = planes.P_47D_30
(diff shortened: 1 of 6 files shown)


rp-:
What crash is the `F_4E_new` causing? and whouldn't it be better to ignore it in the data export script?

Khopa:
Yes, the plane F_4E_new is added to countries.py, but it isn't defined in planes.py iirc.
Yes, I it would be better to ignore it in data export.

rp-:
I'll wait for the update tomorrow and do a export with the F_4E fixes

PR #76, opened 2020-09-26
Added functions to handle multiple DCS installations (to installation.py)

linuxquaker:
Added functions to independently check stable, openbeta, steam installations,
while maintaining compatibility with previous functions.

Tested with my system which installed both stable and openbeta (but not steam),
by adding scripts to if __name__ == "__main__": 

- may need test steam version, which I don't have

### DIFF
--- a/dcs/installation.py
+++ b/dcs/installation.py
@@ -37,65 +37,157 @@ def is_using_dcs_steam_edition():
     except FileNotFoundError as fnfe:
         return False
 
-
-def is_using_dcs_standalone_edition():
+def is_using_dcs_standalone_edition_stable():
     """
-    Check if DCS World standalone edition is installed on this computer
-    :return True if Standalone is installed, False if it is not
+    Check if dcs standalone stable version is installed
     """
-    if not is_windows_os:
-        return False
     try:
         dcs_path_key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Software\\Eagle Dynamics\\DCS World")
         winreg.CloseKey(dcs_path_key)
         return True
     except FileNotFoundError as fnfe:
-        try:
-            dcs_path_key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Software\\Eagle Dynamics\\DCS World OpenBeta")
-            winreg.CloseKey(dcs_path_key)
-            return True
-        except FileNotFoundError:
-            return False
+        return False
 
-
-def get_dcs_install_directory():
+def get_dcs_install_directory_standalone_stable():
     """
-    Get the DCS World install directory for this computer
-    :return DCS World install directory
+        Get dcs stable install dir if avail
     """
-    if is_using_dcs_standalone_edition():
-        try:
-            dcs_path_key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Software\\Eagle Dynamics\\DCS World")
-            path = winreg.QueryValueEx(dcs_path_key, "Path")
-            dcs_dir = path[0] + os.path.sep
-            winreg.CloseKey(dcs_path_key)
-            return dcs_dir
-        except FileNotFoundError as fnfe:
-            try:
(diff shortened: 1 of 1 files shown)


rp-:
I'm currently not at home, to check/test it, but I see that the unit tests failed, can you check that?

linuxquaker:
added is_windows to pass unit test !
- may still needs real env tests especially for steam

rp-:
Thanks for fixing the unittests.
I'm sorry I have to be a bit more picky about this PR, but we should change the error prints to prefix the module e.g.:
`print("dcs-install: Unknown prefer variable %s" % target)`
and also print to stderr (file=sys.stderr), otherwise you might see some output like:
`Unknown prefer variable opnbeta`
and have no clue where this comes from, until you start full text search.

And on other minor thing is the naming, I myself run the openbeta branch in the default "stable" DCS directory, so this might get confusing in the future, but right now I can't come up with anything better.

linuxquaker:
> Thanks for fixing the unittests.
> I'm sorry I have to be a bit more picky about this PR, 

Not at all! I am really glad I can contribute to this project, which I am now working with. Thanks for making pydcs!

> but we should change the error prints to prefix the module e.g.:
> `print("dcs-install: Unknown prefer variable %s" % target)`
> and also print to stderr (file=sys.stderr), otherwise you might see some output like:
> `Unknown prefer variable opnbeta`
> and have no clue where this comes from, until you start full text search.

I totally agree. Just need a little clearing up. the current installation.py code, has some print() s about errors, like:
`print("Couldn't detect DCS World steam installation folder")`
`print("WARNING : Trying to run pydcs on non Windows machine")`

hence i just followed them (doesnt mean that I had a better idea. ^^)
So, do you mean adding module name prefix, and printing to stderr is kind of a new policy we want to apply from now on, to every module, so let's do it when we have a chance?

I am asking because so far I was mainly adding something, not changing previous codes. I can change entire installation.py as you described if that's what you meant!

> And on other minor thing is the naming, I myself run the openbeta branch in the default "stable" DCS directory, so this might get confusing in the future, but right now I can't come up with anything better.

Well, these kinds of problem is the reason I added specific "target" to installation.py

I was in the same boat as you, by changing "stable" to "openbeta" using dcs_updater. at that time, my installation info in winreg was openbeta, but still used /Saved Games/DCS as save folder

A few months later, I reinstalled DCS but this time I used openbeta installer. Then I found that save folder name is changed /Saved Games/DCS.openbeta, so I moved my files to there (this part of my memory is blurry.. I think i manually backed up my saved files)

Another few months later, DCS openbeta became unstable and I changed my install from openbeta to stable, so in this situation install info in winreg shows stable, but saved folder is still using /Saved Games/DCS.openbeta.

That's where pydcs installation.py didnt work for me. (because `get_dcs_saved_games_directory` returned /Saved Games/DCS)

Right now I bought another sdd and running both versions of DCS, stable and openbeta. In this situation pydcs installation.py didnt work either. That's the start of this PR.

It seems quite a few people using both versions, especially after the recent openbeta stability incident. 

Anyway, I understand what you mean. My opinion is that pydcs should not try to magically find out what the user has gone through, hence we need simple but effective policy.

How about,
* if target is stable (or undesignated) , return Saved Games/DCS. 
* if target is openbeta, check if Saved Games/DCS.openbeta exists, return it if true, else return Saved Games/DCS
* if target is steam, return Saved Games/DCS

It's simple but if we checking Saved Games/DCS.openbeta, wouldnt we should also check Saved Games/DCS folder? and that opens a lot of error handling doors. 

Need your input!  Thanks!!

rp-:
> So, do you mean adding module name prefix, and printing to stderr is kind of a new policy we want to apply from now on, to every module, so let's do it when we have a chance?

It must not be the exact module name prefix, but it should be good enough to know where the message comes from, and yes
I think this should be used everywhere. I checked the code and there are not too many prints anyway yet.

> How about,
> if target is stable (or undesignated) , return Saved Games/DCS.
> if target is openbeta, check if Saved Games/DCS.openbeta exists, return it if true, else return Saved Games/DCS
> if target is steam, return Saved Games/DCS

Sounds fine to me

rp-:
closing as there is no progress

PR #80, opened 2020-10-09
Fix callsign export, update data.

DanAlbert:
Most countries don't actually have their callsigns in a table, the
function db.getCallsign will instead refer to either the US or Russia
table depending on the country. Call that instead.

Also fixed a number of nil dereferences that were occurring.


rp-:
Just a small thing, can you add your 3 changed flyable modules to the `flyable` set in `pydcs_export.py:235`, so they always be flyable, as this property depends on the owned modules.

DanAlbert:
Done. Was wondering how that worked :)

rp-:
rebased and merged by hand, thanks

DanAlbert:
Sorry, hadn't noticed the conflict. Thanks!

rp-:
There was no conflict, just wanted to keep the history a bit cleaner.
Nothing to worry.

PR #81, opened 2020-10-15
Added support for scenery destruction zone, and remove scene objects actions

Khopa:
Hello

I added support for two missing mission editor actions : 

* Scenery Destruction Zone
* Scenery Remove Objects Zone

image

Regards,

Khopa


rp-:
Thanks!

Issue #85, opened 2020-10-21
Option to bypass trigger parsing


adamdoda:
Add an option to bypass trigger parsing as it often fails.
Save the original unparsed triggers at load.
When the mission is saved the original triggers are exported.

rp-:
I'm not so sure about this one, I think the target should be to have all triggers/actions parsed correctly.

Do you have missions that fail, so we could add them to unittests and fix those issues?

adamdoda:
I agree that's the ideal way but this an extra option in case parsing fails (it's better to have a fallback than nothing), e.g. we don't need to touch the triggers in https://github.com/UOAF/project-tauntaun so having a reliable load/save is more important than having working trigger import/export. I already implemented this "trigger copier" feature for our app but would be nice to have it in pydcs.

Failing mission: I don't have one now but we met a lot of issues that's why I ended up with this workaround.
Some issues: e.g. the FAC attack group showed up several times and some triggers just gets mixed up if there is lua code in them so load/save "works" but what is exported is invalid.

rp-:
If you ever find such broken trigger stuff again, please post a mission

adamdoda:
More or less nothing is loadable from the DCS site (or at least what I tried).
I'll see if I can dig up the mission for the mixed up logic.

rp-:
You have to note, that some of the missions of the DCS site are veerrry old.
And I had never the scope to be a mission converter or support loading of old missions.
Even DCS does not correctly convert old missions.

adamdoda:
I understand, I have limited time as well, partly that's why I needed this PR instead of the ideal way and fixing the issues.
I was trying newer community missions (and our UOAF missions) and most of the issues are already reported: the now oldest 5-6 one and the weird logic mix up stuff but I have the mission for that and the author allowed me to upload it so I will do it later.

PR #86, opened 2020-10-21
#85 Mission.load_file Bypass triggers option

adamdoda:


rp-:
made a few small changes, but merged

PR #90, opened 2020-10-28
added mod content from files made by pydcs_export.lua

andrewleggett:
at suggestion of the folks over at the Khopa/dcs_liberation discord, I ran pydcs_export.lua with a few mods installed, and this is me sending the files it came out with that had changes against the pydcs files


rp-:
I'm not sure this helps, because:

1. As soon as the next update needs to be exported, your changes are gone(if you are not doing the update yourself again)
2. People that use pydcs with a default dcs, can't use the mod defines anyway, so I'm not sure pydcs is the right place for modded content

So I guess it would make more sense, to put such modded content in some extra modules, that can be included and don't put them directly in the generated base content.

Khopa:
@andrewleggett For DCS Liberation, to support mods, i run the export and extract the diff (manually) to a custom folder. 
I think it is out of scope for pydcs to support modded content. 

It is rather tedious to maintain, but it works, see here : 
https://github.com/Khopa/dcs_liberation/tree/develop/pydcs_extensions

andrewleggett:
Yes, I saw that. I was going to try something along those lines with a new export now the new update is out.

PR #91, opened 2020-11-04
data-export: 2.5.6.57264

Khopa:
Data export for DCS version 2.5.6.57264

It mainly adds a few new units that were added in the WW2 asset pack : 
* M1 37mm AAA
* M45 quadmount AAA
* QF 3.7-inch AAA
* Daimler Armoured Car
* M4 High-Speed Tractor
* Light Tank Mk VII Tetrarch
* FuMG 401 Freya LZ radar

I had to modify the pydcs export script, because of the display name of the QF_17 gun containing the '"' character.


rp-:
Thanks

PR #99, opened 2020-11-18
Fix nearest_airport when coalition is None.

DanAlbert:
When not passing a coalition to nearest_airport, self.airports was being
used exactly rather than only its values.


rp-:
Thanks

PR #100, opened 2020-11-18
data-export 2.5.6.57949

Khopa:
Hello, 

This is the data-export for latest DCS version. (Which include a new F-14A variant)

Khopa


rp-:
Thanks

PR #102, opened 2020-11-22
Add missing entries to `ForcedOptions`. Add types.

DanAlbert:
Adds the following missing options:

* Unrestricted SATNAV
* Wake turbulence
* Cockpit status bar
* Battle damage assessment
* User marks

Adds an enum for the labels option.

Fixes the name of the immortal option, but adds a property to preserve
source compatibility.

Removes the following unused options:

* Scenes
* Tips

I did not preserve source compatibility for the removed options, because
callers that are expecting that to do something should adapt their code.
Even so, writers of these values will still be able to write them.

### DIFF
--- a/dcs/forcedoptions.py
+++ b/dcs/forcedoptions.py
@@ -22,27 +22,47 @@ class ForcedOptions:
         Game = "game"
         Realistic = "realistic"
 
+    class Labels(Enum):
+        None_ = 0
+        Full = 1
+        Abbreviate = 2
+        DotOnly = 3
+
     def __init__(self):
-        self.fuel = None
-        self.easy_radar = None
-        self.mini_hud = None
-        self.accidental_failures = None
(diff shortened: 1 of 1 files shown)


rp-:
I'm not even sure if it is worth to have a compat property, but why not for now.
But we should probably put a comment that this property will soon be removed and people should use the non typo version ;)
Maybe it is even valid to do a `print("don't use this prop anymore", file=std.err)` to warn devs about it.

DanAlbert:
I can just remove it if you prefer. That or a warning is fine with me.

rp-:
Then just remove it, devs should be able to figure out to use the correct one.

DanAlbert:
Done :)

rp-:
Just the build has to succeed ;)

DanAlbert:
Whoops, accidentally mashed some keys I guess. Should be good now.

PR #107, opened 2021-01-07
Add support for new DCS interception range waypoint option

bobmoretti:


bobmoretti:
Tested this with a mission locally; it seems to preserve the interception ranges correctly.

rp-:
Thanks

PR #111, opened 2021-01-28
Point.py | fix error in load_from_dict and added PointAction LandingReFuAr

ugene143:
Added the new PointAction "LandingReFuAr"

Added a check when reading the key "task" from dictionary. This caused problems when using SAM templates.

### DIFF
--- a/dcs/point.py
+++ b/dcs/point.py
@@ -23,6 +23,7 @@ class PointAction(Enum):
     EchelonRight = "EchelonR"
     FromGroundArea = "From Ground Area"
     FromGroundAreaHot = "From Ground Area Hot"
+    LandingReFuAr = "LandingReFuAr"
     Custom = "Custom"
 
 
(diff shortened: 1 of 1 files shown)


rp-:
could you rebase and squash you change to a single commit?
Also `.get()` default default value is None so you don't need to extra state it.

rp-:
hmm something is now totally of with this branch?
Did you force push a wrong branch?

ugene143:
I don't know where I made the mistake. I will create an new branch and PR and we drop this one,ok?

rp-:
I think you should be able to just push the correct commit/HEAD on this branch, but as you feel more comfortable.
I know git can be complicated at some times.
You could create a new branch from master, get the correct commit on that, switch to this branch(`ugene143/dcs`) and `git reset --hard branch-withcorrect-commit`

if you ever lost a commit you can find it with `git reflog` as long git garbage collector didn't run yet.

ugene143:
Thank you for your help.
I think it looks a lot better now^^.

Issue #112, opened 2021-02-07
onboard_num alway is set to 010


rglsphoto1636:
FlyingUnit always get a unit with tail number being set 010, it should be set randomly from 001 to 999 for western  and 01-00 for russian.

rglsphoto1636:
[issue#112.diff.txt](https://github.com/pydcs/dcs/files/5944709/issue.112.diff.txt)
issue fixed, the patch is attached

rp-:
Thanks for the PR and issue report, you are totally right that this isn't handled really well by `pydcs`.
But I would like to have this done correctly at once, this means:

Storing an id/tail-nr map per collation and correctly auto-increment and get numbers from gaps.
This means also on loading the tailnumbers need to be added to the map and on new units new correct numbers can be assigned.
From what I saw from the dcs editor, it starts with `010` and increments by `1` for new planes,
but for helicopters it starts the counter by `050` and even if afterwards a plane is added it will continue at the `05x` level.
So I guess this isn't even well done in the mission editor, but I assume they wanted to have `010`-`049` for planes and
`050`-`???` for helicopters.
For REDFOR this seemed similar?

rglsphoto1636:
Hi RP,

The ME is not handle it well, in a heavy scenario, there would be over one hundred of vehicles in the air, so 010-049 is not enough for fixed planes, also there will be more like in reallife if the number is random than sequencing. 

It's a goog idea for allocating onboard number per collation but this my first look in pydcs and I don't konw how to do it at this moment.

B.R.
Rglsphoto

rp-:
totally understood and still thanks for bringing up this issue.
I'll try to get a PR till the weekend

PR #113, opened 2021-02-08
Fixed onboard number always is 010, now it will be a random number wh…

rglsphoto1636:
…ich is greater than 0 and less than 1000 #112


rp-:
fixed in other PR

PR #119, opened 2021-04-15
Data export 2 7 0 4625

Khopa:
Data export of units for DCS 2.7.0.4625 + terrain update for all maps


rp-:
pushed, but I cleaned up the commit history a bit

Issue #118, opened 2021-04-15
Support for 2.7 weather options


DanAlbert:
Unless you're planning to do this in the next day or so go ahead and assign to me :) Looking to update Liberation to take advantage of this so I'll send a PR this weekend if support hasn't landed yet.

(I haven't looked at this at all yet, so unclear if weather is the only thing that will be needed for 2.7)

rp-:
I just did a quick glance at the mission file and weather section and it seems they really just placed a `clouds` object there with basically the preset, and 2 or 3 other values.
I was expecting the mission file had already the advanced weather stuff, but this looks like the whole thing will be obsolete once we get the real new weather system(in two 2 weeks...).

I'm not sure I have the time to add this, till the weekend, so please go ahead.

DanAlbert:
Sounds good. I'll send you something this weekend if not sooner.

Khopa:
For reference, someone sent this to me : https://www.reddit.com/r/hoggit/comments/mqzv11/hidden_clouds_settings/

There are a few disabled presets, and settings giving an idea of what the future system might be.
But unless we know the miz format, this won't exactly help i guess.

DanAlbert:
> I just did a quick glance at the mission file and weather section and it seems they really just placed a clouds object there with basically the preset, and 2 or 3 other values.

Experimented with a few of these in the ME, and it looks like when using a preset the density is always 0, thickness is always 200, precipitation is always 0 (even for rainy presets), and base is configurable.

I think for now (ED have said that this is going to change again in the future, and it sounds like in the not too distant future) I'll keep it simple and just add the serialization/deserialization for the `preset` field as well as force the other values when serializing. Keeps the API more or less the way it is.

I'll also enumerate all the valid presets so that clients don't need to do that themselves.

DanAlbert:
Although one more detail here is that the configurable base does have a min/max value per preset. It's specified in the file @Khopa linked, so I'll write an extractor for that.

PR #120, opened 2021-04-16
Support DCS 2.7 cloud presets.

DanAlbert:
In 2.7 cloud presets can be selected, which will:

* Override cloud thickness, density, and precipitation
* Bound the cloud base
* Set the "preset" property in the mission file

This patch adds an import script to load the names and cloud base
boundaries from the config file in the DCS install. The script requires
lupa because the parser in dcs.lua is not capable of parsing complex lua
files (I think it's because there are function calls, but I didn't dig
much).

This also removes the `Weather.fog_density` property, since DCS no
longer supports that.

Fixes https://github.com/pydcs/dcs/issues/118


DanAlbert:
I'm relatively confident that this is correct. I've added a handful of unit tests but haven't performed an end-to-end test (generate a mission, load in DCS, see clouds) yet. Will do that soon.

DanAlbert:
Have done an end to end test and it all works but I need to make some changes for this to be 3.6 compatible. Forgot I couldn't use `@dataclass` there.

DanAlbert:
Alright, checks are happy now and this is working for Liberation. Ready to merge as far as I'm concerned.

SnappyComebacks:
I'm not sure dust clouds (just below fog in the ME) are supported either.  Have you seen anything on that?

rp-:
One additional thing I would like to have is an `Enum` with all presets, so I could programmatically say
```
from cloud_preset import Clouds
m.weather.set_cloud_preset(Clouds.Preset3)
```
Or maybe we can even generate names from ui_name. e.g.: Clouds.LightScattered1, Clouds.Overcast3...

rp-:
oh, and I'm not sure if we should put `lupa`, into requierements, because it some kind of soft build dependency.
And a usual user of the library wouldn't need it?

DanAlbert:
> because it some kind of soft build dependency.

It'll end up getting run on CI but it won't be used since we don't run the import scripts in CI. Users will install it as part of their dev environment.

I'll go ahead and remove it and just document the requirement for that script since it's not actually a build dependency like you say. It's unfortunate that requirements.txt can't make these sorts of distinctions. Pipenv can but that might be a difficult change.

> And a usual user of the library wouldn't need it?

I think only the `install_requires` field of the `setup.py` matters for that.

> One additional thing I would like to have is an Enum with all presets

Easy enough. Will add that.

> I'm not sure dust smoke (just below fog in the ME) is supported either. Have you seen anything on that?

The changelog says to not do it _right now_. The ME does not prevent it though. Since pydcs is mostly an API for the ME I don't think we need to prevent that (the ME _does_ prevent setting the cloud base outside the bounds, for example).

Eventually a DCS update will fix the fog/dust options when cloud presets are used and this will just start working without needing a pydcs update.

DanAlbert:
Okay, removed from requirements.txt and added the enum. Should be good to go.

ColinM9991:
Out of curiosity, when are we likely to see this merged and released?

rp-:
@DanAlbert are you fine with this small addition? 67e231923e062afa5e0308f6564e83ae313dbfac

DanAlbert:
@rp- yep, looks good!

rp-:
merged with #125

PR #123, opened 2021-04-18
Respect original data type of onboard_num

### LINKED ISSUE #122: onboard_num failing to parse alphanumeric assignments
`onboard_num` is a string in the mission file but is converted to an integer when interpreted by this module.

This causes the following exception when a mission is loaded that contains a unit with an alphanumerical tail number
>ValueError: invalid literal for int() with base 10: 'E3A'

Is this an intentional design choice?

ColinM9991:
Resolves #122 

Respects the original data type of onboard_num while still allowing uniquely generated tail numbers.

Ran all unit tests before and after making this change with no discrepancies.

I've only recently picked up Python after 7-8 years of years since I last touched it, please do tell me if there's something I haven't done or confirmed.

### DIFF
--- a/dcs/country.py
+++ b/dcs/country.py
@@ -140,6 +140,10 @@ class Country:
             self.current_callsign_category[category] = 0
         return self.callsign.get(category)[self.current_callsign_category[category]]
 
+    @property
+    def unused_onboard_numbers(self) -> Set[str]:
+        return self._tail_numbers
+
     def reset_onboard_numbers(self):
         """
         Resets/clears reserved onboard numbers for this country.
@@ -158,7 +162,7 @@ class Country:
         return is_in
 
     def next_onboard_num(self) -> str:
-        free_set = {str(x) for x in range(10, 999)} - self._tail_numbers
+        free_set = {"{:03}".format(x) for x in range(10, 999)} - self._tail_numbers
         tailnum = free_set.pop()
         self.reserve_onboard_num(tailnum)
         return tailnum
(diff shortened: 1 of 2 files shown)


dcs/country.py
```diff
@@ -140,18 +140,14 @@ def next_callsign_category(self, category):
             self.current_callsign_category[category] = 0
         return self.callsign.get(category)[self.current_callsign_category[category]]
 
-    @property
-    def unused_onboard_numbers(self) -> Set[int]:
```
rp-, line 144: any reason why this property was removed?

dcs/country.py
```diff
@@ -161,8 +157,8 @@ def reserve_onboard_num(self, number: int) -> bool:
         self._tail_numbers.add(number)
         return is_in
 
-    def next_onboard_num(self) -> int:
-        free_set = {x for x in range(10, 999)} - self._tail_numbers
+    def next_onboard_num(self) -> str:
+        free_set = {str(x) for x in range(10, 999)} - self._tail_numbers
```
rp-, line 161: we should here atleast use the same format, as we did for serialization: `"{:03}"`

rp- (changes requested):

dcs/country.py
ColinM9991: I searched high and low for usages and couldn't find any. I can re-add of this is used by consumers.

rp-: well pydcs is a library and so there will be a lot of methods/functions that are potentially not used within the library

ColinM9991: Reintroduced this property

dcs/country.py
ColinM9991: Format has been changed to what it was originally being serialized as

PR #126, opened 2021-04-19
Fixed export script generating wrong weapons display names

Khopa:
Hi @rp- 

I kind of messed up the weapon display name in my latest data-export (https://github.com/pydcs/dcs/pull/119)
I had modified the script to avoid a syntax error, but went too far with safe guarding, so we ended up with underscores instead of spaces in weapon display names, and some chars were even removed.

This PR fixes my mess. 😅 

Khopa


rp-:
also the indentation seems somewhat broken in that `for` loop?
could you have a look?

Khopa:
Hi,

I fixed it, there were some parts of the export file that were indented with tabs, and others with 4-space. 
I reformated the file so we only use 4-space indent (as in the Python code).

(Note : Pycharm lua plugins does not seem to enforce an indent style by default as done on Python file, but i'm using a rather old version, so that might be why.)

rp-:
Thanks

PR #128, opened 2021-04-19
Unit names are not localized in 2.7 fix

adamdoda:


adamdoda:
I tested it with a re-saved old complex mission and a new simple one, neither of them contains unit names in the dictionary so I assume it is the same as for group names.

adamdoda:
I'll fix the tests later.

DanAlbert:
Doesn't this break the ability to load missions that were created with older versions of DCS?

adamdoda:
It will but if we keep the get_name/group_name function it can accidentally hit a key in the dictionary. Actually I just discovered there is a version number in the mission, I don't know how reliable is it though.
Do you have a mission before 2.7, could you check the version?
2.7 version is 19 so that should be 18.
_I have a ~6 month old mission with version 18._

adamdoda:
I will go with 20ce1f3 commit in this PR and the mission version support task needs to be a separate issue/PR.

ColinM9991:
Is it worth holding off on this actually until 2.7 has been merged to the stable branch? This could very well be an issue with 2.7 at present.

DanAlbert:
> It will but if we keep the get_name/group_name function it can accidentally hit a key in the dictionary. Actually I just discovered there is a version number in the mission, I don't know how reliable is it though.
> Do you have a mission before 2.7, could you check the version?
> 2.7 version is 19 so that should be 18.
> _I have a ~6 month old mission with version 18._

I don't think I have one that's newer than that, but I'd trust that it's reliable since it's how the ME itself will be handling this.

rp-:
shouldn't it be that simple to add a:
```
if version < 19:
  # do with translation
else:
  return name
```

into the get_name and we are fine for old and new version? (without thinking about saving right now)

rp-:
And I don't want to support saving in an older format, IMHO we should always write the latest one, otherwise this will be a maintenance hell

adamdoda:
I wouldn't bother with saving either.
I don't know how many of these are out there or how you want to handle it, that's why I created the issue but sure I can just drop an if in there for this one.

DanAlbert:
> And I don't want to support saving in an older format, IMHO we should always write the latest one, otherwise this will be a maintenance hell

Load is all we're worried about from the Liberation side as well, so sounds good.

Issue #134, opened 2021-04-27
Unit Group name throwing AttributeError for DCS 2.7 mission saved in editor


ColinM9991:
Here is a stack trace for an error that occurs when I attempt to save a mission which has previously been saved via the editor for 2.7. It appears the unit name is still an issue.

To reproduce, download this mission and save it via the editor, then try to use pydcs to open edit and re-save.
https://www.digitalcombatsimulator.com/en/files/3313323/

```
Traceback (most recent call last):
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\discord\ext\commands\core.py", line 85, in wrapped
    ret = await coro(*args, **kwargs)
  File "<server_path>\Desktop\Bot\cogs\dcsmissioncog.py", line 47, in set_weather
    dcs_mission.save()
  File "<server_path>\Desktop\Bot\helpers\dcsmissioneditor.py", line 107, in save
    self.mission.save()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\mission.py", line 1907, in save
    zipf.writestr('mission', str(self))
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\mission.py", line 1979, in __str__
    return lua.dumps(self.dict(), "mission", 1)
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\mission.py", line 1956, in dict
    m["coalition"][col] = self.coalition[col].dict()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\coalition.py", line 291, in dict
    d["country"][i] = self.country(country).dict()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\country.py", line 194, in dict
    d["plane"]["group"][i] = plane_group.dict()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\unitgroup.py", line 514, in dict
    d = super(FlyingGroup, self).dict()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\unitgroup.py", line 253, in dict
    d = super(MovingGroup, self).dict()
  File "<server_path>\AppData\Roaming\Python\Python39\site-packages\dcs\unitgroup.py", line 183, in dict
    "name": self.name.id,
AttributeError: 'str' object has no attribute 'id'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #136, opened 2021-04-28
fix format 19 problems and always save in current mission format

rp-:
So @ColinM9991 noticed in #134  that there are still some problems with loading/saving.
And I finally got some time to really think about the issue.
I noticed, pydcs since some time, always stored missions, if not specified either in version `16` or the version of the loaded mission,
IMHO this isn't very good anymore, as we already have plenty of new mission properties and so on, so we should always write the latest supported version number.
This also means we shouldn't store group/unit/waypoint names anymore as translatable strings as this was removed in `2.7`.
I'm fine with correctly loading old missions files, but we should always try to write them in the latest DCS format.

Opinions @DanAlbert, @Khopa, @adamdoda  ?


DanAlbert:
Haven't reviewed the change itself but +1 to the theory. Generating things as closely as possible to what the ME produces is helpful for debugging.

adamdoda:
+1

ColinM9991:
I've just tested this with my Discord bot and can confirm that it works as expected and without issue.

I think the idea of fixing the mission version to the latest supported version is a good idea. At the very least it means this codebase doesn't grow into a monolith.

PR #138, opened 2021-05-03
Add properties and color fields to TriggerZone

SnappyComebacks:
TriggerZones in new missions always have a `properties` field and `color` field.  I've added both to the TriggerZone constructor.

In order to get the tests to pass, I had to add the `properties` field to every TriggerZone in the test `.miz` files.  Otherwise a parse method would error when it did not find `['properties']`.


dcs/triggers.py
```diff
@@ -8,13 +8,14 @@
 
 
 class TriggerZone:
-    def __init__(self, _id, position: mapping.Point, radius=1500, hidden=False, name=""):
+    def __init__(self, _id, position: mapping.Point, radius=1500, hidden=False, name="", color={1: 1, 2: 1, 3: 1, 4: 0.15}, properties={}):
```
rp-, line 11: You have to be very careful with default dicts/lists/sets/... as this doesn't behave as one would think in python.
In the end every TriggerZone object with a default dict would use the same ref to that dict.
So the usual approach here is to give the default param `None` and check in the `__init__` if it is `None` and create than a default with the intended values.

dcs/triggers.py
```diff
@@ -48,15 +50,16 @@ def load_from_dict(self, data):
...
+                imp_zone["name"],
+                imp_zone["color"],
+                imp_zone["properties"]
             )
-            tz.color = imp_zone["color"]
             self._zones.append(tz)
             self.current_zone_id = max(self.current_zone_id, tz.id)
 
-    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="") -> TriggerZone:
+    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="", color={1: 1, 2: 1, 3: 1, 4: 0.15}, properites={}) -> TriggerZone:
```
rp-, line 60: Same as in `__init__` don't create dicts in default parameters

dcs/triggers.py
```diff
@@ -48,15 +50,16 @@ def load_from_dict(self, data):
                 mapping.Point(imp_zone["x"], imp_zone["y"]),
                 imp_zone["radius"],
                 imp_zone["hidden"],
-                imp_zone["name"]
+                imp_zone["name"],
+                imp_zone["color"],
+                imp_zone["properties"]
```
rp-, line 55: As you mentioned, that you have to fix the .miz files, this isn't how it should work.
pydcs should try to load any old mission file and you shouldn't need to manually fix them.
So here as in other `load_from_dict` methods either check if `properties` is in the import dict or use `imp_zone.get("properties", {})`

rp- (changes requested):
So thanks for your afford, please see the review comments.
And revert all unittest .miz files as they are supposed to work, as they are.

dcs/triggers.py
SnappyComebacks: I think I have made the change as you said to.  Let me know if I should have done something different.

SnappyComebacks:
I have reverted the tests and they all pass now after taking your suggestions.

dcs/triggers.py
```diff
@@ -48,15 +55,22 @@ def load_from_dict(self, data):
...
+                imp_zone["name"],
+                imp_zone["color"],
+                imp_zone.get("properties",{})
             )
-            tz.color = imp_zone["color"]
             self._zones.append(tz)
             self.current_zone_id = max(self.current_zone_id, tz.id)
 
-    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="") -> TriggerZone:
+    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="", color=None, properites={}) -> TriggerZone:
```
SnappyComebacks, line 65: Looks like I spelled `properties` wrong here.

SnappyComebacks: @rp- , does this also need to be `properties=None` and follow the same pattern as what I had done with color in `__init__`?

rp-: yes same problem

dcs/triggers.py
```diff
@@ -8,13 +8,19 @@
...
 
 class TriggerZone:
-    def __init__(self, _id, position: mapping.Point, radius=1500, hidden=False, name=""):
+    def __init__(self, _id, position: mapping.Point, radius=1500, hidden=False, name="", color=None, properties={}):
         self.id = _id
         self.radius = radius
         self.position = copy.copy(position)
         self.hidden = hidden
         self.name = name
-        self.color = {1: 1, 2: 1, 3: 1, 4: 0.15}  # TODO color attributes
```
rp-, line 17: You can simplify this even more:
```
self.color = color if color is not None else {1: 1, 2: 1, 3: 1, 4: 0.15}
```

Also you forgot properties:
```
self.properties = properties if properties is not None else {}
```

rp- (changes requested):

SnappyComebacks:
Should be good to go now.  Let me know if I missed something.

dcs/triggers.py
SnappyComebacks: Done.

dcs/triggers.py
```diff
@@ -48,15 +50,20 @@ def load_from_dict(self, data):
...
-            tz.color = imp_zone["color"]
             self._zones.append(tz)
             self.current_zone_id = max(self.current_zone_id, tz.id)
 
-    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="") -> TriggerZone:
+    def add_triggerzone(self, position: mapping.Point, radius=1500, hidden=False, name="", color=None, properties=None) -> TriggerZone:
         self.current_zone_id += 1
-        tz = TriggerZone(self.current_zone_id, position, radius, hidden, name)
+
+        real_color = color if color is not None else {1: 1, 2: 1, 3: 1, 4: 0.15}
```
rp-, line 63: sorry kinda my fault here, we actually don't need defaults at all.
just pass the parameters to the TriggerZone constructor
defaults will be set anyway in the constructor than if they are None:
```
tz = TriggerZone(self.current_zone_id, position, radius, hidden, name, color, properties)
```

rp- (changes requested):

rp-:
And also take a look at the flake errors

dcs/triggers.py
SnappyComebacks: I receive `SyntaxError: non-default argument follows default argument` when removing the defaults from `color` and `properties`.  `hidden` and `name` are the culprits here.

rp-: This works for me?
```
    def add_triggerzone(
            self,
            position: mapping.Point,
            radius=1500, hidden=False,
            name="",
            color=None,
            properties=None
            ) -> TriggerZone:

        self.current_zone_id += 1

        tz = TriggerZone(self.current_zone_id, position, radius, hidden, name, color, properties)
        self._zones.append(tz)
        return tz
```

SnappyComebacks: Oh I misunderstood what you were saying.  I thought you were talking about the TriggerZone constructor itself.  I just pushed a commit to mimic what you have above.

rp-:
if you could fix the final flake8 errors it is good to merge

SnappyComebacks:
@rp- Is the flake8 config your builds are using available somewhere?  I could only see the "max-line-length" one.  I would like to match configs so we don't have to run over and over for flake8 specifically.

rp-:
Flake commands are in: `dcs/.github/workflows/pythonpackage.yml`

SnappyComebacks:
Should be good now (hopefully).

PR #142, opened 2021-05-09
Added SetCallsign command

gronank:
Added added the basic SetCallsign command and put it in the wrappedactions dict. Don't know if there's a list of ground-unit-only commands this shoulod be entered in too somewhere.


gronank:
@rp- linter should be happy now, though I can't seem to be able to run the pipeline myself to verify.

PR #146, opened 2021-05-23
Fix payload bug, improve interface and lookup.

DanAlbert:
There was a bug where the max key was being computed based on the max
key of the *input* dict rather than of the dict being modified, so if
the new loadout had an index equal to the current index + 1, we'd
clobber the existing (potentially non-matching) loadout.

This doesn't actually need to be an integer indexed dict though, as this
is never used directly; it's a data structure internal to the class. As
such, I replaced the max index tracking with a name -> payload dict.
This also improves the name based loadout lookup.

The second commit adds an API for injecting payload directories.

We've been doing this in Liberation by manually appending to
`FlyingType.payload_dirs`, but that's an implementation detail that we
ought to not touch.

This adds a real API for configuring both fallback (when the directory
should be the least preferred loadout source) and preferred (when the
directory should be used over any other option) loadout directories.

I did also remove the pydcs-specific loadout directory scanning
(dcs/payloads) because no such things exists.


rp-:
Manually merged, a small request, if you have such small flake8 fixes, could you squash/rebase them and force push.
I really want to keep the git history somewhat clean.

DanAlbert:
Yeah, can do. GitHub makes longer reviews annoying if you force push (can't use "changes since my last review") so I got out of the habit. I prefer the cleaner history same as you though :)

rp-:
Yeah, I think it is fine after a review, to just push the review fixes and then if its good to rebase.
Thanks

PR #157, opened 2021-06-20
Fix weapon names in exporter.

DanAlbert:
These are supposed to be human readable names *with* spaces and special
characters, so shouldn't be escaped aside from double quotes.

The latest export seemingly also has a bunch of new liveries and pylon
data.


DanAlbert:
Any issues here?

rp-:
no sorry, I'm just currently on vacation and barely on the PC :D

DanAlbert:
That's a good reason, sorry for disturbing you :) Enjoy!

Issue #160, opened 2021-07-08
enforced typing on all APIs?


DanAlbert:
I can do the work here if that's preferred (and I assume it is), but I wanted to make sure that the patch would be welcome first.

We (DCS Liberation) hit a not small number of bugs where we have type mismatches between our code and pydcs. mypy does its best for libraries without complete type info, but it's pretty common for anything imported that isn't sufficiently typed to end up degrading to `Any`, at which point mypy does nothing.

What I'd like to do is send a (probably quite large, but also should be quite straightforward) PR that adds any missing type annotations and enforces `disallow_untyped_defs = True` in mypy.ini, and make sure that runs as part of the github actions.

Any objections? Type annotations can take a bit of work to maintain, but IME it's quite small, especially when compared to the number of bugs they catch.

rp-:
yes please, go ahead!

DanAlbert:
https://github.com/pydcs/dcs/pull/161 addresses the first two todo items.

PR #161, opened 2021-07-09
Clean up existing mypy issues, run in CI.

DanAlbert:
This is the first step for https://github.com/pydcs/dcs/issues/160.

Each commit here was already squashed, so should be independently reviewable (and can rebase the PR when submitting and it won't make a mess of the branch history).

There are four commits here:

1. Fixes/deletes some files that were just plain broken in the current repo. Not all the vehicle/ship/etc class names were fixed after one of the recent DCS exports. I deleted one since it's presumably unused, but I fixed templates.py since it was used elsewhere. I can resurrect the deleted file if preferred, but it seems no one uses it because it's been broken for a while.
2. Latest export _before_ mypy changes. I needed to tweak a few things in the exporter to add missing annotations and fix a bug or two, and this commit makes it easier to see the result of those later changes.
3. The interesting part: cleans up all the mypy errors that exist for the default(ish) configuration. I did punt on the terrain files since that exporter also needs to be fixed, and it's a lot more annoying to run those so I haven't gotten to them yet. For now they're disabled in the mypy.ini, but no other mypy configuration is used beyond the default yet.
4. Run mypy as part of the GitHub action to prevent regressions.

Most of the fixes needed were pretty straightforward. New base classes, correcting existing annotations, adding missing annotations, introducing a few generics, etc. A very small number of APIs changed, and mostly only to correct the signature to the *actual* behavior (a handful of functions treat all arguments as optional even though if that behavior is relied upon the function will raise). Shouldn't be any serious impact from the changes.

Many of the function-internal issues were fixed by adding an assertion or a type-check that raises since the existing APIs can't be made type-safe without being more invasive, and IMO that's better left until *after* mypy is fully deployed. It might also break existing users.


DanAlbert:
For whatever reason my first attempt at disabling the checking for the terrain files worked on my machine but not in CI. Fixed now and ready to merge as far as I'm concerned. I'll be following up with the other items mentioned in the bug.

mypy.ini
```diff
@@ -0,0 +1,24 @@
...
+ignore_errors = True
+
+[mypy-dcs.terrain.persiangulf]
+ignore_errors = True
+
+[mypy-dcs.terrain.syria]
+ignore_errors = True
+
+[mypy-dcs.terrain.thechannel]
+ignore_errors = True
```
rp-, line 24: super minor, but newline is missing here at the end of file :)

DanAlbert: Whoops, thanks. Fixed.

Issue #162, opened 2021-07-09
drop Python 3.6 support?


DanAlbert:
I mistakenly use something from 3.7+ just about every time I make a change here. Any idea if there are any users still depending on 3.6 support? For us (Liberation), all we care about is 3.9 (and in a few months we'll only care about 3.10), but idk what other users need.

Supporting 3.6 keeps us from using (among other, less interesting things):

* `from __future__ import annotations` (allows using non-quoted type names in type annotations for circular imports, factory classmethods, comparators, etc)
* dataclasses

Dropping 3.6 is the only really interesting one (to me at least) at the moment since those two features are the ones I always expect to be able to use until the github checks say no :) 3.9 is nice because it lets you use `list[...]` instead of importing `List` from `typing` (and same for all the other built in types), but that's pretty minor.

If there are users that still depend on 3.6 support this isn't a big deal so can just be closed, but if not it'd be nice to move forward, expecially since 3.10 will bring some basic type contracts that can presumably improve some of the APIs.

rp-:
I'm pretty sure everyon could easily upgrade to 3.7 or even 3.9.
So I have no objections to go to 3.7

PR #163, opened 2021-07-09
Drop Python 3.6 support. Test Python 3.9.

DanAlbert:
Fixes https://github.com/pydcs/dcs/issues/162.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


DanAlbert:
(If you want to let this one sit for a while to see if anyone speaks up on behalf of 3.6 that's fine with me, no urgent need for this)

rp-:
Please update also this here:
https://github.com/pydcs/dcs/blob/2b4eb8bf99d225b37305735773d6504ad844d144/setup.py#L29

DanAlbert:
> Please update also this here:
> https://github.com/pydcs/dcs/blob/2b4eb8bf99d225b37305735773d6504ad844d144/setup.py#L29

Good catch. Done.

rp-:
guess we can drop the change in pythonpackage.yml now?

DanAlbert:
> guess we can drop the change in pythonpackage.yml now?

Huh, I don't remember making the change that was submitted. Not sure what happened there.

I've rebased. I think what I have here is _currently_ the same (always test with the latest python 3 available on in CI), but once 3.10 is available this will silently change behavior, so I've made it explicit.

PR #164, opened 2021-07-09
Move flake8 config into setup.cfg.

DanAlbert:
This way the same configuration used in the GitHub actions can be easily
used locally.


.github/workflows/pythonpackage.yml
```diff
@@ -26,16 +26,10 @@ jobs:
...
-        flake8 . --count --max-complexity=10 --max-line-length=127 --statistics --ignore=C901,W503 \
-          --extend-exclude=tests/,doc/,tools/,dcs/helicopters.py,dcs/planes.py,dcs/statics.py,dcs/ships.py,dcs/weapons_data.py,dcs/vehicles.py \
-          --per-file-ignores="__init__.py:F401 syria.py:E126 thechannel.py:E126 persiangulf.py:E126 normandy.py:E126 nevada.py:E126 caucasus.py:E126 marianaislands.py:E126 dcs/lua/test_parse.py:E501"
-    - name: Type checking
-      run: |
-        pip install mypy
-        mypy dcs
+        # Check the default configuration.
+        flake8 .
+        # But also run a limited number of checks to verify that the excluded files are at least correct syntax.
```
DanAlbert, line 31: Pretty sure I understood the intent here. Once mypy is more completely rolled out this can be dropped since it'll be checking the generated files as well.

PR #166, opened 2021-07-09
Fix incorrect type annotations, add py.typed.

DanAlbert:
These commits together are the errors I found while updating Liberation to type check our uses of pydcs. In one case there was an actual bug (loading a mission with a refuel/rearm landing waypoint and then resaving it would corrupt the mission by writing a string back where an integer was needed).

This also adds the py.typed file, without which mypy ignore type annotations in this package when checking dependents: https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

### DIFF
--- a/dcs/mission.py
+++ b/dcs/mission.py
@@ -14,7 +14,7 @@ from pathlib import Path
 from typing import Any, List, Dict, Sequence, Union, Optional, Type
 
 from dcs.coalition import Coalition
-from dcs.terrain.terrain import Warehouses
+from dcs.terrain.terrain import Terrain, Warehouses
 from dcs.triggers import Triggers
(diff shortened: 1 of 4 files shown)


DanAlbert:
If you'd prefer one big patch lmk. I'm working through getting Liberation up to the latest and finding a lot of little things.

rp-:
hmm maybe one PR with multiple commits, not sure its useful to have 1 PR per commit, until something depends on it.

DanAlbert:
Can do. Will finish up my migration and then send them in a batch. I'm done for the day though so you have a reprieve :)

DanAlbert:
Okay, ready to go. Checks another box on #160.

PR #168, opened 2021-07-15
Use floats for qnh and temperature

magwo:
From testing through Liberation, DCS appears to accept float values just fine.
If mmHg is an integer, there is poor resolution. For example the inches Hg value of 29.94 would not be possible to have. Next up from 29.92 would be 29.96.


magwo:
For the record: If, in the mission editor in DCS 2.7, I input 21.2330001 for temp, and 29.93 for QNH, it shows up as 
```
["temperature"] = 21.233,
["qnh"] = 760.222,
```
in the mission file.

PR #173, opened 2021-07-30
Raise TypeError when not possible to set radio channels

magwo:
Instead of silently returning, when clearly the operation did not complete with success.

### DIFF
--- a/dcs/flyingunit.py
+++ b/dcs/flyingunit.py
@@ -125,11 +125,11 @@ class FlyingUnit(Unit):
              frequency: The frequency to set the channel to, in megahertz.
 
         Raises:
+            TypeError: Unit type has no configurable radios, or it's unsupported.
             KeyError: No such radio or channel number.
         """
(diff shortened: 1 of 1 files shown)


dcs/flyingunit.py
```diff
@@ -125,11 +125,11 @@ def set_radio_channel_preset(self, radio_id: int, channel: int,
              frequency: The frequency to set the channel to, in megahertz.
 
         Raises:
+            TypeError: Unit type has no configurable radios, or it's unsupported.
             KeyError: No such radio or channel number.
         """
         if self.radio is None:
-            # Not all aircraft have configurable radio channels.
-            return
+            raise TypeError(f"{self.type} doesn't have configurable radio channels")
```
DanAlbert, line 132: `FlyingGroup.set_frequency` needs to be fixed to suppress this. `set_frequency` is called for setting the AI's channel, but it also sets channel 1 for the given radio.

PR #174, opened 2021-07-31
Dont fail so soon when p graph files are not found

magwo:
The failure on p file loading was causing issues when building a python project with pydcs into an exe file. This PR solves it.

Considering that the city_graph objects are not widely used in pydcs, I think it would be preferrable to fail only when the p files are actually needed. That way you can use pydcs as a library without the p files, as long as you don't use the destroy_oil_transport.py code.

### DIFF
--- a/dcs/terrain/caucasus.py
+++ b/dcs/terrain/caucasus.py
@@ -3024,7 +3024,7 @@ class Caucasus(Terrain):
     center = {"lat": 43.69666, "long": 32.96}
     bounds = mapping.Rectangle(380 * 1000, -560 * 1000, -600 * 1000, 1130 * 1000)
     map_view_default = MapView(mapping.Point(-255714.28571428, 680571.42857143), 1000000)
-    city_graph = Graph.from_pickle(os.path.join(os.path.dirname(__file__), 'caucasus.p'))  # type: Graph
+    city_graph = None
     temperature = [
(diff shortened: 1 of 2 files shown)


magwo:
More generally I feel that the randomized mission generation stuff should be moved into a separate library that uses pydcs.

I think it would be neater if pydcs was fully focused on loading, editing, creating and saving miz files.
The content-creation parts of random missions could be an external software, perhaps under the same GitHub organization.

rp-:
Well those scripts should not really give you a full fledged super dynamic generated mission, but should be more seen as an example, how to use pydcs.

Anyhow, please in the future try to cleanup your merge request, when you think it is done. I want to have a somewhat clean version history and now unnecessary commits(like Fix lint error).

magwo:
> Well those scripts should not really give you a full fledged super dynamic generated mission, but should be more seen as an example, how to use pydcs.
> 
> Anyhow, please in the future try to cleanup your merge request, when you think it is done. I want to have a somewhat clean version history and now unnecessary commits(like Fix lint error).

Sure! Was thinking we would squash the commits like we usually do in liberation, but both ways works for me.

rp-:
It is fine to add commits, after a review was done, so one could see what was done after the review.
But before and after that I would welcome it, if the branch is already cleaned up.
Thanks!

PR #176, opened 2021-08-01
Prevent failures when setting radio presets by checking for radio

### LINKED ISSUE #175: can't set AI comm frequency for aircraft without preset channel support
https://github.com/pydcs/dcs/pull/173 introduced a bug in `FlyingGroup.set_frequency` for any aircraft without preset channel support. That function is used to set the AI's comm frequency, but it also tries to set channel 1 to the same frequency. That function now raises when used for aircraft without preset channel support. The new error needs to be suppressed there.

magwo:
Closes #175

### DIFF
--- a/dcs/flyingunit.py
+++ b/dcs/flyingunit.py
@@ -97,7 +97,8 @@ class FlyingUnit(Unit):
 
     def set_default_preset_channel(self, freq: float) -> None:
         """Sets the frequency for channel 1 of the main radio."""
-        self.set_radio_channel_preset(1, 1, freq)
+        if self.radio:
+            self.set_radio_channel_preset(1, 1, freq)
 
     def set_radio_preset(self):
         """Resets the radio channel configuration to the airframe's default."""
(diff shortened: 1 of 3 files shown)


magwo:
Should it perhaps be `radio is not None`?

dcs/flyingunit.py
```diff
@@ -97,6 +97,8 @@ def reset_loadout(self):
 
     def set_default_preset_channel(self, freq: float) -> None:
         """Sets the frequency for channel 1 of the main radio."""
+        if self.radio is None:
+            return
```
DanAlbert, line 101: This is for setting channels, so probably should raise, and the callers should have the check?

magwo: Sure!

tests/test_mission.py
```diff
@@ -506,7 +506,8 @@ def test_radio_channels(self):
         unit.set_client()
         frequency = 300.0
         self.assertNotAlmostEqual(unit.radio[2]["channels"][1], frequency)
-        unit.set_radio_channel_preset(2, 1, frequency)
```
rp-, line 509: Do we really need to fix the test here?

magwo: Not really. My thinking was that if someone changes the type for the test to an A-10 it will still work,

Also, if the test code sometimes is used as an example for other code, maybe it's good to show that you need to check for `radio` before setting channel presets?

PR #177, opened 2021-08-18
Don't validate cloud base - only warn

magwo:
When editing (opening, modifying and saving) an existing mission using pydcs, it doesn't really make sense to validate the mission. Doing so will cause problems with editing existing missions that for some reason deviate from pydcs's idea of what is valid.

For example, a mission produced directly with the Mission Editor had a cloud base of 800. Because of this it was not possible to re-save the mission with pydcs. The Mission Editor appears to be more liberal than pydcs.

If we want to raise validation errors, that should probably be done when _setting_ the property during the editing phase - not in the `save` phase.
unknown

### DIFF
--- a/dcs/weather.py
+++ b/dcs/weather.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import logging
 import random
 import math
 from enum import Enum
@@ -50,7 +51,7 @@ class CloudPreset:
 
     def validate_base(self, base: int) -> None:
         if not self.min_base <= base <= self.max_base:
-            raise ValueError(
+            logging.warn(
                 f"Cloud base {base}m is out of range for {self.name}. Must be between "
                 f"{self.min_base}m and {self.max_base}m.")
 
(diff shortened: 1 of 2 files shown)


DanAlbert:
> The Mission Editor appears to be more liberal than pydcs.

The cloud base values come directly from DCS: https://github.com/pydcs/dcs/blob/master/tools/weather_import.py. Do we have the units wrong, is the data just out of date and needs to be re-exported, or is DCS's data wrong?

I don't think removing the validation is the way to go. Fixing the check is. We rely pretty heavily on pydcs telling us when we're wrong, and warnings are extremely easy to miss. pydcs isn't doing anyone any favors by silently generating a mission that it believes is broken. Removing the check here just means that every downstream project needs to duplicate it themselves.

> If we want to raise validation errors, that should probably be done when setting the property during the editing phase - not in the save phase.

Agreed.

magwo:
Ok agreed, I will see if I can move the validation stuff to the edit phase.

PR #178, opened 2021-08-19
Add A-10C presets support, with tests

magwo:


magwo:
Confirmed working by an A-10C II flying friend.

dcs/a10c_presets.py
```diff
@@ -0,0 +1,88 @@
...
+            16: 261000000,
+            17: 267000000,
+            18: 251000000,
+            19: 253000000,
+            20: 266000000,
+        }
+
+    def load_dict(self, data):
+        presets = data.get("settings", {}).get("presets", {})
+        for chan, freq in presets.items():
```
magwo, line 31: I guess this is a bit superflous? Could we just assign self.presets to the dict from the data?

magwo:
Discovered this now: https://forums.eagle.ru/topic/226274-a10c-radio-presets-overwrite-the-viper-presets/

Was able to reproduce - the Viper presets are indeed overwritten by the presence of this UHF_RADIO/SETTINGS.lua file in the miz.

I don't fully understand the situation. It seems we can produce these SETTINGS.lua files for each individual aircraft now?

magwo:
Ok I now have a better understanding of the A-10C presets, and how their architecture recently changed.

This PR will change course and add support for airframe-individual presets.

dcs/mission.py
```diff
@@ -2,6 +2,7 @@
 The mission module is the entry point to all pydcs functions.
 """
 import copy
+
```
rp-, line 5: why the newline?

tests/test_mission.py
```diff
@@ -734,3 +734,25 @@ def test_mission_save_pictureFileName(self):
...
+        print('-' * 10, "Saved", save_path)
+        # reload mission
+        m = dcs.mission.Mission()
+        msgs = m.load_file(save_path)
+        self.assertEqual(0, len(msgs))
+
+        self.assertIsNotNone(m.a10c_presets)
+        self.assertEqual(m.get_10c_preset(1), 251.0)
+        self.assertEqual(m.get_10c_preset(2), 252.0)
+        self.assertEqual(m.get_10c_preset(20), 253.0)
```
rp-, line 758: newline missing here

dcs/mission.py
```diff
@@ -254,6 +257,11 @@ def loaddict(fname: str, mizfile: zipfile.ZipFile, reserved_files: List[str]) ->
                 mapresource_dict = loaddict('l10n/DEFAULT/mapResource', miz, reserved_files)
                 self.map_resource.load_from_dict(mapresource_dict, miz)
 
+            if 'UHF_RADIO/SETTINGS.lua' in miz.namelist():
```
rp-, line 260: so is this SETTINGS.lua file always bound to the a10c presets?
I really don't know just asking if there could be more than just a10 preset within this file?

dcs/a10c_presets.py
rp-: should work, until we need/want a deep copy

rp- (changes requested):

dcs/mission.py
magwo: Thanks for reviewing!

My understanding is it used to be A-10C and KA-50 only, until recently. See this comment I made on the PR:

> Ok I now have a better understanding of the A-10C presets, and how their architecture recently changed. This PR will change course and add support for airframe-individual presets.

Consider this PR paused for now. It will look different when it is done for individual airframes. It will still be these SETTINGS.lua files, but they will be in airframe-id subdirectories in the miz file.

dcs/mission.py
magwo: Not sure, I ran flake8 and fixed until it stopped complaining. ;)

PR #183, opened 2021-09-19
Use manual heading for ground spawned aircraft

magwo:
Ground-spawned aircraft can have their heading set, and if pydcs doesn't respect that it will rewrite the heading to point towards the first waypoint, when editing a mission file.

I'm not sure if this is the most elegant way to do this. I don't fully understand the code. These spawn types are not defined in any constants in the code.

Before open/save:

heading_before

After saving with pydcs (first aircraft has been manually corrected):

heading

### DIFF
--- a/dcs/unitgroup.py
+++ b/dcs/unitgroup.py
@@ -181,6 +181,9 @@ class Group(Generic[UnitT, PointT]):
         for u in self.units:
             u.skill = skill
 
+    def is_ground_spawned(self) -> bool:
+        return False
+
     def dict(self):
         if not isinstance(self.name, str):
             raise TypeError("Point name expected to be `str`")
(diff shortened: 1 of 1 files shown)


magwo:
Will fix the linter formatting error, but in the meantime I would appreciate a discussion on the approach.

rp-:
so just so I understand, you want to have the set unit heading, if the unit has no waypoint?

What I ask myself now is, what heading or better from where does a coldstart unit get the heading?

magwo:
> so just so I understand, you want to have the set unit heading, if the unit has no waypoint?
> 
> What I ask myself now is, what heading or better from where does a coldstart unit get the heading?

No. There's a fairly new concept in DCS where you can set airplanes to "Takeoff from ground", meaning they spawn on the ground, not using a specific slot on the airfield.

And when they are set like this, you can set the heading of the aircraft in the Mission Editor. Pydcs needs to use this manually set heading, and not overwrite it with the heading-towards-first-waypoint logic currently in use.

Untitled

magwo:
Just tested my use case with the this commit hash installed. Works as intended - fixes my problem.
I'm using it in this project: https://github.com/magwo/dcs-mission-buzzer
Mostly an internal software we use in Master Arms to randomize weather and more, within the parameters we like.
Our standard training mission has some units that use Takeoff From Ground, and they get rotated in the random-weather mission file ouput by the tool.

Ready for merge if it looks ok.

magwo:
Thanks!

PR #186, opened 2021-10-14
Draw tool support

magwo:
Draw Tool support.


magwo:
Confirmed that pydcs can currently open missions containing drawings. It's just that they will not be included when the mission is re-saved by pydcs.

tests/test_drawings.py
```diff
@@ -0,0 +1,118 @@
...
+        red_layer = m.drawings.get_layer(StandardLayer.Red)
+        red_layer.add_icon(Point(1000, 1000), StandardIcon.MechanizedArtillery)
+        mission_path = 'missions/New_mission_w_added_std_icon.miz'
+        m.save(mission_path)
+
+        m2 = dcs.mission.Mission()
+        self.assertEqual(0, len(m2.load_file(mission_path)))
+        red_layer = m.drawings.get_layer(StandardLayer.Red)
+        
+        self.assertEqual(StandardIcon.MechanizedArtillery.value, red_layer.objects[0].file)
```
rp-, line 118: missing newline

rp- (changes requested):
sorry took me a while, was on vacation the last two weeks.
could you also rebase/fixup the "Correct Mistakes" commit (for nicer history)
Than it looks fine for me

tests/test_drawings.py
magwo: I would squash that commit, but it appears the merge of master I did into this branch would get lost/overwritten by picks if I did that... and I'm worried it will result in weirdness/conflicts when that is merged to master. 🤔 😬 

So I'd rather not, I think. Will fix the newline.

rp-:
manually merged with a few history cleanups

rp-:
Thanks!

magwo:
Thanks for reviewing and merging!
I've used this stuff in a less serious side project and the API is holding up well I think. These racetracks are made programmatically.
racetracks

PR #190, opened 2021-11-15
Add Disembarking task

adamdoda:
Add missing Disembarking task

### DIFF
--- a/dcs/task.py
+++ b/dcs/task.py
@@ -843,6 +843,27 @@ class Embarking(Task):
             "distribution": {x: {y: y for y in distribution[x]} for x in distribution}
         }
 
+class Disembarking(Task):
+    """Unload task for helicopters.
+
+    Helicopter group will land at the given coordinates and will unload the given groups.
+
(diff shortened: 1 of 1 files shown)


adamdoda:
Manually tested with project-tauntaun load/save works, I didn't test if they actually do the task in game but I assume, looks like a pretty simple task.

dcs/task.py
```diff
@@ -844,6 +844,28 @@ def __init__(self, position: Point = Point(0, 0), groupids: List[int] = None,
...
+    """Unload task for helicopters.
+
+    Helicopter group will land at the given coordinates and will unload the given groups.
+
+    :param position: :py:class:`dcs.mapping.Point` where to land and unload
+    :param groupids: list of groups to unload
+    """
+    Id = "Disembarking"
+
+    def __init__(self, position: Point = Point(0, 0), groupids: List[int] = None):
```
rp-, line 857: Sorry for being picky, but it should be:
```suggestion
    def __init__(self, position: Point = Point(0, 0), groupids: Optional[List[int]] = None):
```

PR #195, opened 2022-01-01
Add MapObjectIsDead TriggerRule Condition

RndName:
This adds the c_dead_zone triggerrule condition which allows to check if a mapobject was killed. Plan is to use this triggerrule to handle killed map objects in dcs liberation.

With the DCS Patch 2.7 ED renamed the UI Name for the condition within the MissionEditor from `DEAD_ZONE` to `MAP OBJECT IS DEAD`. Thats why i gave the class the name MapObjectIsDead instead of DeadZone which is not really meaningful

Source: https://www.reddit.com/r/hoggit/comments/mrc8oy/psa_map_objects_now_have_dedicated_trigger/

### DIFF
--- a/dcs/condition.py
+++ b/dcs/condition.py
@@ -541,6 +541,27 @@ class IndicationTextEqual(Condition):
         return d
 
 
+class MapObjectIsDead(Condition):
+    predicate = "c_dead_zone"
+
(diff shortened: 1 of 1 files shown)


RndName:
fixed the lint error. :(

rp-:
Thanks

PR #196, opened 2022-01-06
Terrain update (from DCS 2.7.9.18080)

RndName:
Caucasus, Mariana, Nevada and PG had some updates since the last export

Biggest change here is the rename of nearly all Nevada Airports.

I do not own the WW2 maps so i cant not create an export for them.


RndName:
needs some updates to fix the problems with the renamed bases on nevada map. im on it

PR #198, opened 2022-01-12
Adding OptFormation Data

SnappyComebacks:
I feel like I took an unconventional approach to this when compared to all the other Option Tasks.

DCS has a different formation format for rotary aircraft than fixed wing aircraft.  Some formations have extra options, some do not.

Because there are three different values for each formation, I chose to use static methods be the main method of generating these formations.

Wasn't sure if it was possible/how to test this in the existing tests, so I only tested it by incorporating it into DCS Liberation and checking if the values of the formations I used looked right.

Let me know if you disagree with stuff.  I wasn't sure if I should put `airplane` or `fixed_wing` in front of the fixed wing only options like I did with `rotary`.

This is the mission I made to source the data (changed extension to .zip): 
[FormationDefinitions.zip](https://github.com/pydcs/dcs/files/7852232/FormationDefinitions.zip)

### DIFF
--- a/dcs/task.py
+++ b/dcs/task.py
@@ -1660,7 +1660,8 @@ class OptFormation(Option):
         WW2_FighterVic = 14
 
     class VariantIndex(IntEnum):
-        """The distance that the formation is spread apart is dependent on the formation index.  Ignore text for rotaries, its always small-medium-large."""
+        """The distance that the formation is spread apart is dependent on the formation index.
+        Ignore text for rotaries, its always small-medium-large."""
         Close = 1
         Open = 2
         GroupClose = 3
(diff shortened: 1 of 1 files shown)


SnappyComebacks:
I see the failing test(s).  I wonder if I have to have default values for the new variables?

dcs/task.py
```diff
@@ -1588,12 +1588,459 @@ def __init__(self, value: Values = Values.UseWhenFlyingInSAMWEZ):
...
+        """Only used by rotaries.  Combines with FormationIndex to vary left/right."""
+        Right = 0
+        Left = 1
+
+    def __init__(
+        self,
+        value: Values,
+        formationIndex: FormationIndex,
+        variantIndex: Optional[VariantIndex],
+        zInverse: Optional[ZInverse]
```
rp-, line 1679: yes, these params should have defaults.
Because the task will be created with a simple `cls()` call, and afterwards just their `params` will be overwritten

rp- (changes requested):

dcs/task.py
```diff
@@ -1588,12 +1588,459 @@ def __init__(self, value: Values = Values.UseWhenFlyingInSAMWEZ):
...
 
 class OptFormation(Option):
+    """Formations changed data structures during DCS development.  Use the static methods to build the desired formation."""
+
     Key = 5
 
-    # TODO missing option enums
+    formationIndex: int
+    variantIndex: int
+    zInverse: int
```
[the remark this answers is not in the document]
SnappyComebacks: Oops thanks.  Was definitely on autopilot when I wrote this part.  =)

dcs/task.py
SnappyComebacks: Thanks!  I have added defaults for an Echelon Right Open formation.

Issue #17, opened 2019-03-27
Position Point to Lat Long?


JohanAberg:
Hi,
Thanks for this great project.

Is there a way to convert a Mission unit's position to Lat/Long values?
I'd like to plot unit's position on a real map. 
Any ideas or pointers please?

Cheers

rp-:
There is currently no function in the library for that.
But it is possible, I just didn't bother to implement or add an dependency to [pyproj4](https://github.com/pyproj4/pyproj).
With that library it should be fairly easy if we pick the correct projection method,
I think hints for that can be found in the terrian .lua config files in DCS.

JohanAberg:
Great thanks, I'll take a closer look at pyproj.

DanAlbert:
It seems the projection parameters for the Caucasus are available (`Caucasus.MapCoords.lua`), but not for the others. Any idea if that's something that could be extracted from DCS like all the other data? I've poked around a little but haven't found anything so far.

rp-:
No, I have no idea.
I'm afraid this is hidden in the native code somewhere for newer modules

DanAlbert:
I finally got around to finishing this. https://github.com/Khopa/dcs_liberation/commit/8a01209ded308af429e825e48b23e3c1f9138505

@rp- do you want this in pydcs? It would require a dependency on pyproj.

rp-:
I'm fine with that, as a lot of other users could profit from that and don't have to reinvent the wheel themselves

DanAlbert:
Still planning to do this but haven't found the time just yet. There are a few choices in API though that I'd like your thoughts on, @rp-.

The complication is that any point <-> LL conversion is map dependent, so any call to convert needs to know which map is being used.

Using this in Liberation, I added the conversion methods on our Theater object. It's okay, but it does mean plumbing the map information around anywhere we need to convert.

Option 2 (my preference) is to subclass Point into MapPoint which would have the map data and conversion methods. Any cases that don't need conversion could be more cheaply and easily constructed, but we'd still have type safe calls to the conversions.

Is there an option 3? Global state is no good because the API allows multiple missions to exist simultaneously. Precomputing has all the same problems of needing context, but also forces the cost when it's not needed.

DanAlbert:
> From your experience, which Points you don't need to convert?

[...]

PR #203, opened 2022-02-21
Add latitude/longitude APIs to Point.

DanAlbert:
This is more or less the code we've been using in Liberation for months, but made more convenient by adding the APIs directly to `Point`, and not requiring the caller to keep a reference to the `Terrain` to convert a `Point` to a `LatLng`. Since those transformations are map specific, `Point` now requires a reference to `Terrain`.

I haven't done any caching of the computed `LatLng` for `Point`. I could add that in one of two ways:

1. Make `Point` a frozen dataclass
2. Make `x` and `y` `@property` so we can invalidate the cache if they are mutated.

I prefer option 1 if we decide caching is needed, but I'd prefer to land this as-is and revisit if needed. We haven't been doing any caching of these in Liberation and profiling suggests it is a very small cost, and the `LatLng` is used in few enough places that caching would not improve much anyway. Construction of the transformers is expensive, but those are cached in the `Terrain` so they're only constructed once.

The inverse of this operation, `Point.from_latlng`, _does_ require the caller to pass the `Terrain` explicitly rather than storing the `Terrain` in `LatLng`. In our experience the `LatLng` being converted to a `Point` is usually from an external source anyway and constructed in-situ, so the caller likely has a reference to `Terrain` anyway.

For the few cases where Point was being used as a relative offset (in which case the terrain does not matter), I've introduced `Vector2`.

A few tasks here took a default `Point(0, 0)`. I'd be surprised if those worked before. I've made those non-defaults since there no longer is a default constructible Point.

Conveniently (for everyone but me) this required some changes in all the airport data, so this includes a fresh export of all the terrain data.

Fixes https://github.com/pydcs/dcs/issues/17.


DanAlbert:
flake8 should be happy now. I've disabled it in the terrain implementation files since those are largely auto-generated. If you'd prefer I run an autoformatter on them lmk which one to use.

I should mention that I haven't adapted Liberation to this yet. We have a few bespoke constructed points there, so I'll have a better idea of how annoying this might be once I've done that. I suspect most of our uses are either secretly `Vector2`, or we already have the `Terrain` nearby, since an arbitrary `Point` isn't much use.

DanAlbert:
(am still working on the tests, apparently I didn't run _all_ of them)

DanAlbert:
Okay, flake and test clean now. Next I'll probably find out that I accidentally used something not in one of the older Python versions :)

DanAlbert:
Made the `Point` math operation accept a `Vector2` for the RHS so offsets don't need terrain references.

PR #206, opened 2022-02-22
Make Terrain picklable again.

DanAlbert:
The transformers are not pickleable. Make them cached properties so they
aren't serialized.

### DIFF
--- a/dcs/terrain/terrain.py
+++ b/dcs/terrain/terrain.py
@@ -446,6 +446,7 @@ class Terrain:
         map_view_default: MapView,
     ) -> None:
         self.name = name
+        self.projection_parameters = projection_parameters
         self.bounds = bounds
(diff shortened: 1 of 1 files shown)


DanAlbert:
Struggling to make the caching compatible with 3.7. I guess for now I'll just do it manually.

Thoughts on when/if you want to drop 3.7 support? I don't think we have nearly as much reason as we had for dropping 3.6, but if none of your users are on 3.7 it doesn't really matter.

rp-:
Then keep the better version somewhere around and we could apply it if we drop python 3.7 support.

DanAlbert:
> Then keep the better version somewhere around and we could apply it if we drop python 3.7 support.

Sounds good. If you ever decide to do that, feel free to tag me to do it :)

Issue #214, opened 2022-03-18
Vehicle 'transportable' property absent during import


spencershepard:
This occurred after adding vehicles in the mission editor with 2.7.11.21408.1

From the DCS 2.7.11.21408.1 Open Beta changelog:
"ME. Mission editor does not read Transportable boolean when opening .miz file - fixed."

My error:
Traceback (most recent call last): File "C:\RotorOps\Generator\MissionGenerator.py", line 208, in scenarioChanged source_mission.load_file(filename) File "c:\rotorops\generator\venv\src\pydcs\dcs\mission.py", line 376, in load_file status += self.coalition[col_name].load_from_dict(self, imp_mission["coalition"][col_name]) File "c:\rotorops\generator\venv\src\pydcs\dcs\coalition.py", line 114, in load_from_dict unit.load_from_dict(imp_unit) File "c:\rotorops\generator\venv\src\pydcs\dcs\unit.py", line 78, in load_from_dict self.transportable = d["transportable"] KeyError: 'transportable'

rp-:
Hi!

It would always be good, if you could attach such a mission for testing

spencershepard:
[transportable error miz.zip](https://github.com/pydcs/dcs/files/8302191/transportable.error.miz.zip)

Hi there, okay yes no problem.  Attached is an original miz, created before the DCS update.  I opened this in the mission editor, latest DCS version after today's update, and saved it as a new file.  This one shows the error I described.  Also attached.  

Thanks!

DanAlbert:
If you have the time to test with https://github.com/pydcs/dcs/pull/215, I can merge that. I haven't been able to test it much yet but it seems plausible.

spencershepard:
It seems to solve my issues from a quick test.  Thanks guys

PR #215, opened 2022-03-18
Remove the transportable property from units.

DanAlbert:
It seems this property is no longer used in the latest miz version.

Fixes https://github.com/pydcs/dcs/issues/214


DanAlbert:
(I haven't tested this much yet)

PR #217, opened 2022-04-06
Update AttackGroup task with missing params

RndName:
AttackGroup was missing some params which were already added to other tasks. So i copy pasted the params from the AttackUnit task while checking which DCS adds by default and whats the default value. You can see the task object created from the mission editor with just default settings:
grafik

So the changes match them.
Also add the missing "All" value for the Expand Param

### DIFF
--- a/dcs/task.py
+++ b/dcs/task.py
@@ -323,24 +323,29 @@ class AttackGroup(Task):
 
     :param group_id: group id to attack
     :param weapon_type: :py:class:`WeaponType` to use for the attack
+    :param attack_limit: how much attack runs
+    :param group_attack: indicate if the unit must be attacked by all aircrafts in the group.
+    :param altitude: altitude used for attack
+    :param direction: direction of attack
+    :param expend: how many ammunition to expend
     """
     Id = "AttackGroup"
 
-    def __init__(self, group_id=0, weapon_type: WeaponType = WeaponType.Auto,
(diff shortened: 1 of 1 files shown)


dcs/task.py
```diff
@@ -316,23 +326,25 @@ class AttackGroup(Task):
     """
     Id = "AttackGroup"
 
-    def __init__(self, group_id=0, weapon_type: WeaponType = WeaponType.Auto):
+    def __init__(self, group_id=0, weapon_type: WeaponType = WeaponType.Auto,
+                 attack_limit: Optional[int] = None, group_attack=False,
+                 altitude_enabled=False, altitude=0, direction_enabled=False,
```
DanAlbert, line 331: Replace `altitude_enabled: bool = False` and `altitude: int = 0` with `altitude: Optional[int] = None`? Ditto for `direction`.

RndName: you're right. i just copy pasted it without going into it deeper.
I adopted the parameters for AttackGroup. Left the other ones untouched as this would affect ppl using this Task alread with the parameters.

I also set the Default values for all Attack Tasks to 0 or 1 depending on the parameter instead of adding None to the actuall mission file inside the miz. DCS has not complained about it but its a wrong data type for it so i think they already have some error handling which takes care of it. But to be on the safe site and future proof i aded the correct default values.

PR #229, opened 2022-05-29
Add debug logging to the export script.

DanAlbert:
There are a handful of (commented out) prints scattered through the file
bug I can't seem to get print or any of the other usual DCS lua logging
mechanisms to log anywhere. Work around that by writing our own log
file.

I haven't converted much, just the parts that I'm currently debugging.


DanAlbert:
@rp-: Requesting your review rather than pushing this myself in case this is redundant and I just failed to find the right way to do this.

.gitignore
```diff
@@ -6,3 +6,4 @@ dist/
 pydcs.egg-info/
 doc/_*
 venv/
+/export.log
```
rp-, line 9:
```suggestion
/export.log

```
missing NL

rp- (changes requested):
just the missing NL, else LGTM

PR #232, opened 2022-06-16
Add drawing package to setup.py (missed in #231)

bobmoretti:
In #231 I added the `terrain.projections` package to `setup.py`. Unfortunately I missed the fact that there is a `drawing` package that needs to be added as well. This PR adds that to the `package` list in `setup.py` as well.


bobmoretti:
I feel really stupid for missing this one. Sorry for the bother.

bobmoretti:
Thanks!

Issue #230, opened 2022-06-13
terrain export instructions no longer work


DanAlbert:
[These instructions](https://github.com/pydcs/dcs/issues/36#issuecomment-640748908) no longer work for the latest DCS. I tried doing this for the new Falklands map and got the following error:

```
2022-06-13 01:43:36.729 ALERT   LUACOMMON: Error: GUI Error: [string "./MissionEditor/modules/me_map_window.lua"]:4514: attempt to index global 'MapWindow' (a nil value)
GUI debug.traceback: stack traceback:
	[C]: ?
	[string "./MissionEditor/modules/me_map_window.lua"]:4514: in function 'dumpairportdata'
	[string "./MissionEditor/modules/me_map_window.lua"]:4543: in function 'createAircraft'
	[string "./MissionEditor/modules/me_map_window.lua"]:2270: in function 'onMouseDown'
	[string "./dxgui/bind/NewMapView.lua"]:362: in function 'callback'
	[string "./dxgui/bind/Widget.lua"]:299: in function <[string "./dxgui/bind/Widget.lua"]:298>
```

Khopa:
Hey @DanAlbert ,

If you copied pasted, the  "dumpairportdata" function which is provided the top of the issue, I think this version does not work anymore. 

I don't have access to my DCS PC to test right now, but, IIRC, you just have to change ```MapWindow``` to ```base.MapWindow```, like in I did in the screenshoot from my latest comment, and it should work. ED just changed their import/module structure at some point.

Khopa

DanAlbert:
Interesting. I've done this a few times and never had to do that before. I'll try again at some point.

PR #234, opened 2022-07-03
Add support for Falklands terrain

Hornet2041:
adds support for the Falklands terrain. Followed @Khopa 's thread from another post.

resolves https://github.com/pydcs/dcs/issues/230


dcs/terrain/falklands.py
```diff
@@ -0,0 +1,668 @@
...
+    center = {"lat": 52.468, "long": 59.173}
+    bounds = mapping.Rectangle(74967, -114995, -129982, 129991)
+    map_view_default = MapView(mapping.Point(0, 0), 1000000)
+    city_graph = None
+    temperature = [
+        (5, 17),
+        (5, 17),
+        (2, 14),
+        (2, 14),
+        (2, 14,
```
DanAlbert, line 621: The check is failing because you missed the closing parenthesis here.

dcs/terrain/falklands.py
```diff
@@ -0,0 +1,668 @@
...
+        (-5, 11),
+        (1, 15),
+        (1, 15),
+        (1, 15),
+        (5, 17)
+    ]
+    assert (len(temperature) == 12)
+
+    def __init__(self):
+        super(Falklands, self).__init__("falklands")
```
DanAlbert, line 633: Did the tool generate this? This doesn't match the constructor for `Terrain` any more. See the other terrains for examples. One of the things you'll notice is that you need projection parameters so we can do lat/on <-> x/y conversions. https://github.com/pydcs/dcs/blob/master/tools/export_map_projection.py is the tool to do that. See the docs at the top.

DanAlbert (changes requested):

DanAlbert (approved):
I pushed an update that fixes the issues mentioned. Also a fresh export since there are a couple new airports.

DanAlbert:
But I need to figure out why mypy is angry at this map and only this map... (also I forgot to update the projections import)

PR #236, opened 2022-07-09
Livery scanner

Raffson:
I'm trying to replace the livery-export with something that scans the available liveries at run-time, as this was suggested by @DanAlbert in the following issue: https://github.com/dcs-liberation/dcs_liberation/issues/1861
What I currently have should work properly. Though I'd like to know whether it would be interesting if I introduce liveries for ground units as well. I believe a vast majority of ground units is already present in the livery paths that get scanned, so it should be a matter of a couple adjustments to make this happen.


Raffson:
Back to draft since I realized I still need to export the short names for countries because that's what's being used in the description.lua files and so that's also how it's stored in the "map".

Raffson:
Closing PR since I noticed I still have to take care of default liveries, this will require some extra adjustments and while at it I'll make sure I can integrate my branch in liberation before re-opening this PR

Raffson:
After careful testing with @Starfire13, I believe this is ready for review.

@rp- Should I also include an "export"? Because these changes do require regenerating `planes.py` and `helicopters.py`, and I would assume the other files are necessary to guarantee correct behavior. Alternatively I could try editing those files manually with a regex or something, but that seems way more tedious than simply exporting data from DCS.

rp-:
without having a look at the whole change.
I see this actually a 2 different things, the current liveries, where just meant to be a static picture of a somewhat "current" DCS installation and should provide mission coders help in picking correct liveries with auto completion.

But your change is more a dynamic livery scanner, where a runtime user can pick it's liveries from whatever he has installed.
IMHO this should probably 2 different things, the old one could probably kept in place at is it is, and the dynamic one should be it's one function calls.

Raffson:
> I see this actually a 2 different things, the current liveries, where just meant to be a static picture of a somewhat "current" DCS installation and should provide mission coders help in picking correct liveries with auto completion.

I can understand wanting to have the autocomplete feature, but when I start weighing the pros and cons I'm not sure if it's worth it.
The first thing I see as a problem is the fact that there's custom liveries present in the current version of pydcs. So there's already no guarantee it'll work as expected. Then there's the fact that, specifically in planes.py, 75% of the code are just definitions for the Liveries, corresponding with roughly 30k lines of code. The combination of those two already seem to make it worth to least consider ditching the current livery system. On top of that, if someone wants to write a mission using pydcs, they can always refer to their installation/saved games folder to see which liveries they can use. But nevertheless, I still think you've got a point with the whole autocomplete thing...

Raffson:
@rp- I'm looking at some options to include the autocompletion and by the looks of it I think it's doable. Would you consider the scanner as a replacement given a working autocomplete feature?

rp-:
You have some valid points, so I'm also fine with a version without autocomplete, because probably no one is using it anyway ;)

Raffson:
I've been looking at this:
https://stackoverflow.com/questions/63436496/is-it-possible-to-have-python-ides-offer-autocompletion-for-dynamically-generate
 And that:
https://stackoverflow.com/questions/13603088/python-dynamic-help-and-autocomplete-generation/13603392#13603392

@rp- Seems like PyCharm doesn't want to play ball unless I generate some stub files. Does that sound good to you?

rp-:
I'm also fine with a version without it

Raffson:
> I'm also fine with a version without it

Are you sure?

I currently half-way into a stub-generator. The idea is to allow the following things to happen:

````
print(Liveries.A_10A)
print(Liveries.F_16C_50.x14th_Fighter_Squadron)
print(Liveries.F_16C_50.x14th_Fighter_Squadron.name)
print(Liveries.map[Liveries.LEOPARD_2.livery_folder_name])
````

The first example would return the set of available liveries, which currently works with the stub that I have.

The other 3 are still work in progress, none of them work except for the fact that I get the auto-completion suggestions. I assume the only thing I need to do is make a separate container (a class basically) for the set of liveries in order to add the liveries as attributes, because python doesn't want me adding attributes to a 'set'-object.

As for how the stub generator works, very simple, the user should just run `liveries_scanner.py` once in order to generate the stub. I don't see another way since liveries may vary across users, so letting the user generate the stub would make more sense imo, essentially making the auto-complete feature optional. So what I did is make a template that I put in a separate directory, and use that to generate the rest of the stub. The actual stub is added to .gitignore for obvious reasons.

If this doesn't work out by the end of the day, I'll just prepare for a merge without it. But that still brings me back to whether I should include a data export of DCS or not...

Raffson:
Well, I think I got it... The following actually works, with auto-completion:

````
print(Liveries.A_10A)
print(Liveries.F_16C_50.x14th_Fighter_Squadron)
print(Liveries.F_16C_50.x14th_Fighter_Squadron.order)
print(Liveries.F_16C_50.x14th_Fighter_Squadron.countries)
print(Liveries.map[Liveries.LEOPARD_2.unit_livery_id])
````

Raffson:
I got everything sorted out, except for PyCharm generating warning for certain things:

image

image

Despite the warning the code actually does what's expected, and at the end of the day I can't see why people would want to manually add liveries since the whole point of the code is scanning the liveries for you. The warnings w.r.t. the loops can be solved by simply adding `.__iter__()`, however then you get another "unfilled" warning. It's definitely not ideal, but it does provide you with auto-completion to make life a lot easier. If you prefer I can still revert the commit and drop this whole 'stub'-thing...

Raffson:
After several attempts to get the build to work, I'm concluding I'll have to include an export. I'll have to carefully examine all changes before I can make another attempt at passing the checks...

Raffson:
Ah, finally it checks out 🤣 

With the release of the Mirage F-1, this PR is coming right in time. I would guess the liveries for those aircraft would've easily doubled the lines of code in `planes.py` since they added a whole bunch of variants.

Raffson:
Won't be resolving these conflicts until #242 is merged, otherwise I end up exporting twice for no particular reason...

Raffson:
@rp- What's your opinion on the addition of 'X_' to identifiers that start with a number? Currently it just adds an underscore but as I pointed out, this prevents PyCharm's auto-completion from suggesting those options.

Raffson:
Also, here's the list of changes I went through, aside from the needed adjustments in the export script and the addition of liveries_scanner.py:

- Adding "shortname" for countries, since those are the ones being used in the description lua files of the liveries.
- Adding "X_" instead of an underscore which prevents auto-completion from making all suggestions. I would think this is a reasonable approach, however if you're not convinced just say the word and I'll pull this change out.
- Changes in "unittype.py" w.r.t. the addition of the livery scanner.
- Single change in "flyingunit.py", a transitive result of the changes in "unittype.py" because we're now using a country's "shortname".
- Changes in "installation.py", mainly because we want to make a better attempt at determining the saved games folder. If this folder doesn't get determined correctly, custom liveries won't get processed. The user will be forced to re-initialize the livery scanner which could be a problem when having designs like `planes.py` and `helicopters.py`. If the livery scanner isn't initialized on time, `None` will be returned instead and when using class attributes as done in the mentioned files, those modules will have to be reloaded, which can be a real pain. Therefore I can stress enough how important it is for the livery scanner to be initialized on time.

rp-:
fine to merge from my side, or is there still something you like to add?

Raffson:
> fine to merge from my side, or is there still something you like to add?

I think I should run another export, just to be sure no 'flyable' attributes were missed. I expect to do this somewhere in the next hour or 2.

Aside from that I think everything's been covered 👍

Raffson:
@rp- Double checked with a new export and I get no changes, meaning all 'flyable' attributes are correct. Did a manual inspection to verify this and everything checks out, so I think it's ready for merge 👍

PR #238, opened 2022-07-18
Remove all mentions of `meters` from scenery actions

### LINKED ISSUE #207: Remove Scene Objects doesn't work
It seems that the attribute "meters" is no longer an attribute for this action (in old missions it was, however it didn't make much sense, since the zone itself already has a radius). So new missions using this can't be parsed, since pydcs expects the attribute to be present.

```
Attempting to load mission file C:/Users/DCSServerUser/Saved Games/DCS.openbeta_server/Missions/MasterArms_Training_CA_FullMap_ClearWeather.miz
  File "main.py", line 33, in <module>
  File "dcs\mission.py", line 355, in load_file
  File "dcs\triggers.py", line 211, in load_from_dict
  File "dcs\triggers.py", line 116, in create_from_dict
  File "dcs\action.py", line 1724, in create_from_dict
KeyError: 'meters'
[6920] Failed to execute script 'main' due to unhandled exception!
```

### LINKED ISSUE #237: KeyError when importing modules with scenery removal/destruction zones
In the latest master (`f263a3`), with a mission saved from the current version of DCS openbeta (2.7.15.26783), I get the following error loading into PyDCS:

```
  File "f:\projects\dcs\pydcs\dcs\mission.py", line 355, in load_file
    self.triggerrules.load_from_dict(self, imp_mission["trigrules"])
  File "f:\projects\dcs\pydcs\dcs\triggers.py", line 217, in load_from_dict
    self.triggers.append(trigger_map[d[x]["predicate"]].create_from_dict(mission, d[x]))
  File "f:\projects\dcs\pydcs\dcs\triggers.py", line 122, in create_from_dict
    action_ = action.actions_map[actions[a]["predicate"]].create_from_dict(actions[a], mission)
  File "f:\projects\dcs\pydcs\dcs\action.py", line 1753, in create_from_dict
    return cls(d["destruction_level"], d["zone"], d["meters"])
KeyError: 'meters'
```
I didn't investigate this fully, but it looks like the `meters` table key was removed from DCS at some point. I tried creating circular and polygonal trigger zones, and ran into the same issue.

In `actions.py`, there is some clue as to what's going on here

```python
        # Note : the parameter meters is in the DCS save file, but seems unused, and not editable from ME
        self.meters = meters
        self.params.append(self.meters)
```
I wonder if ED recently went through and cleaned up this (apparently unused) parameter.

In any case, it's trivial to remove any reference to `meters` from the two actions. I will submit a PR unless someone knows a good reason not to (compatibility with non-openbeta version?)

Miz file that repros attached as .zip.
[scenetest.zip](https://github.com/pydcs/dcs/files/9128942/scenetest.zip)

bobmoretti:
This attribute seems to be unused in prior versions of DCS, and now no longer added by the mission editor.

Should fix #207.

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -1713,50 +1713,42 @@ class RemoveSceneObjectsMask(IntEnum):
 class RemoveSceneObjects(Action):
     predicate = "a_remove_scene_objects"
 
-    def __init__(self, objects_mask: RemoveSceneObjectsMask = RemoveSceneObjectsMask.ALL, zone="", meters=1000):
+    def __init__(self, objects_mask: RemoveSceneObjectsMask = RemoveSceneObjectsMask.ALL, zone=""):
         super(RemoveSceneObjects, self).__init__(RemoveSceneObjects.predicate)
         self.objects_mask = objects_mask
         self.params.append(self.objects_mask)
         self.zone = zone
         self.params.append(self.zone)
-        # Note : the parameter meters is in the DCS save file, but seems unused, and not editable from ME
-        self.meters = meters
-        self.params.append(self.meters)
 
     @classmethod
     def create_from_dict(cls, d, mission):
-        return cls(RemoveSceneObjectsMask(d["objects_mask"]), d["zone"], d["meters"])
+        return cls(RemoveSceneObjectsMask(d["objects_mask"]), d["zone"])
 
     def dict(self):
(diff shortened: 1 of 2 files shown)


bobmoretti:
I also checked that pydcs still loads (legacy) mission files with the `meters` attribute correctly/without errors.

Gentle ping to @rp- to take a look at this when you have a moment :).

PR #241, opened 2022-07-22
Export didn't include Mi-24P as flyable

Raffson:
For some reason the export doesn't recognize the Mi-24P as flyable on my end. Maybe because I don't have the module installed?

One thing's for sure, it should be flyable...

### DIFF
--- a/dcs/helicopters.py
+++ b/dcs/helicopters.py
@@ -2729,6 +2729,7 @@ class AH_64D_BLK_II(HelicopterType):
 
 class Mi_24P(HelicopterType):
     id = "Mi-24P"
+    flyable = True
     height = 4.354
     width = 17.3
     length = 20.953


Khopa:
Hi @Raffson,

You should include Mi-24P in this list from the export script, so it not manually required to change it next time data is exported :

See:
https://github.com/pydcs/dcs/blob/7a220848e8db7e02455ede0903d763a47d972915/tools/pydcs_export.lua#L305-L310


Khopa

Raffson:
> Hi @Raffson,
> 
> 
> 
> You should include Mi-24P in this list from the export script, so it not manually required to change it next time data is exported :
> 
> 
> 
> See:
> 
> https://github.com/pydcs/dcs/blob/7a220848e8db7e02455ede0903d763a47d972915/tools/pydcs_export.lua#L305-L310
> 
> 
> 
> 
> 
> Khopa

Right, don't know how I missed this 😂

But still I'm wondering, how did other people manage to correctly export the Mi-24P without your suggested change? Is it because they have it installed or...?

Either way, I'll give it a shot and verify it exports correctly, if it does I'll make a small PR with your suggestion, and while at it I assume I should also add the F1 to that list.

Raffson:
@Khopa As I suspected, the Mirage F1CE has the same problem, another PR coming up! 👍

Khopa:
>Right, don't know how I missed this 😂

No worry, you can't really know that until you run into the issue. I made the mistake many times. 😁 

See: https://github.com/pydcs/dcs/blob/7a220848e8db7e02455ede0903d763a47d972915/tools/pydcs_export.lua#L377-L379

By default it should work, but it seems that the `HumanCockpit` property isn't always available in the export we get from dcs, so there is the `flyable` list as a fallback for almost half the flyable aircraft.

Raffson:
> By default it should work, but it seems that the `HumanCockpit` property isn't always available in the export we get from dcs, so there is the `flyable` list as a fallback for almost half the flyable aircraft.

Made that a full list now 😃

PR #242, opened 2022-07-23
Fix Flyables

Raffson:
Re-organized the flyables by type (jet engine/piston engine/helicopter) which makes verification a little easier, and sorted them alphabetically. Ran another export to be sure.

Two things I'd like to elaborate:

- Most of the additions in the flyables map don't have any effect, since it was already being exported correctly, I simply added them just to guarantee it stays exporting the flyable types correctly.
- I changed `F/A-18C` into `FA-18C_hornet`, since the former is referring to the AI model. Either way, the hornet (flyable one) was already being exported correctly so this simply removes the "flyable" option for the AI version.

### DIFF
--- a/dcs/planes.py
+++ b/dcs/planes.py
@@ -381,7 +381,6 @@ class F_A_18A(PlaneType):
 
 class F_A_18C(PlaneType):
     id = "F/A-18C"
-    flyable = True
     height = 4.66
     width = 11.43
     length = 17.07
@@ -16855,6 +16854,7 @@ class H_6J(PlaneType):
 
 class Christen_Eagle_II(PlaneType):
     id = "Christen Eagle II"
+    flyable = True
     height = 1.9812
     width = 11.594846
     length = 5.6388
@@ -28215,6 +28215,7 @@ class Hawk(PlaneType):
 
 class I_16(PlaneType):
     id = "I-16"
+    flyable = True
     height = 3.25
     width = 9.004
     length = 6.13
@@ -37266,6 +37267,7 @@ class MiG_21Bis(PlaneType):
 
 class Mirage_F1CE(PlaneType):
     id = "Mirage-F1CE"
+    flyable = True
     height = 4.5
     width = 8.4
     length = 15.3
--- a/tools/pydcs_export.lua
+++ b/tools/pydcs_export.lua
@@ -304,11 +304,11 @@ file:close()
 -------------------------------------------------------------------------------
 local flyable = {}
 -- Jet engine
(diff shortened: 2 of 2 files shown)


dcs/planes.py
```diff
@@ -381,7 +381,6 @@ class Pylon9:
 
 class F_A_18C(PlaneType):
     id = "F/A-18C"
-    flyable = True
```
rp-, line 384: should this be removed?

Raffson: > should this be removed?

I'm pretty sure it should, it's the regular C-variant which is AI only afaik. The flyable F-18's ID should be "FA-18C_hornet".

DanAlbert:
> Most of the additions in the flyables map don't have any effect, since it was already being exported correctly, I simply added them just to guarantee it stays exporting the flyable types correctly.

How will we know if the auto-detection is working now?

Raffson:
> How will we know if the auto-detection is working now?

Imo I wouldn't count on auto-detection since it has already proven it can fail. I think maintaining a map of flyable aircraft is more sensible, and it's not like DCS released an aircraft every 5 minutes so...

So in other words, I sort of consider auto-detection to be "broken" / "incomplete". Fixing it would definitely be interesting, however it'll require careful inspection of the modules that fail to see if they have something in common that got missed.

Either way, I currently have my hands full with other PRs so I don't see myself doing this in the next few days...

Raffson:
> How will we know if the auto-detection is working now?

It turned out to be pretty straight forward. The auto-detection has a prerequisite, i.e. the export has to run on a system with all aircraft modules installed, planes & helicopters. The export script which uses `plane.HumanCockpit` for its auto-detection will always return `nil` if aircraft in question isn't installed.

I concluded this because of the following observations:

The first export PR I opened was before I had the F1 installed, and after realizing the Mi-24 didn't export properly, it didn't take too long until I also realized the same applied to the F1. I then changed the flyables map, resulting in a correct export but the latter is not the actual clue because I forced it in a sense. However I did some debugging tweaks to check which aircraft have a "HumanCockpit" that's not nil, and that's where I noticed it. The F1CE was showing up as non-nil, which wasn't the case before I installed the F1, otherwise the export that missed the Mi-24 would've included the F1CE, which it didn't.
Comparing the list of flyables that was present before my changes to the modules that I've got installed, and then compare those to the aircraft that got missed, it's an exact match so I guess that's case closed.

Until we can find another property that guarantees a correct answer, I think maintaining the map of flyable aircraft is the best way to go.

PR #246, opened 2022-08-15
Catch and log unicode errors in liveries.

DanAlbert:
The code looks right (it's doing utf-8 decoding where expected), but a
user is reporting an issue that seems to be a wrongly encoded livery. We
can safely skip these, but log them so it's clear why they aren't
available.

https://github.com/dcs-liberation/dcs_liberation/issues/2368

### DIFF
--- a/dcs/liveries_scanner.py
+++ b/dcs/liveries_scanner.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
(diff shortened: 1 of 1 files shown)


DanAlbert:
(leaving as a draft until I have said user confirm that this works)

PR #248, opened 2022-08-15
Support ANSI description files in livery scanner

Raffson:
https://github.com/dcs-liberation/dcs_liberation/issues/2368

Reverts commit 2a7230db48dd21495d05d52c905b88cb2b34b883 in the process.

### DIFF
--- a/dcs/liveries_scanner.py
+++ b/dcs/liveries_scanner.py
@@ -65,10 +65,10 @@ def generate_stub() -> None:
 
 class Livery:
     id: str = ""
-    """ID of the livery, corresponds with the folder-name of the livery."""
+    """ID of the livery, corresponds with the folder-name of the livery. To be used in mission file!"""
 
     name: str = ""
-    """Name of the livery"""
+    """Name of the livery as displayed in DCS"""
 
     order: int = 0
     """Order of the livery used to sort like DCS"""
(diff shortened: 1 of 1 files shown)


dcs/liveries_scanner.py
```diff
@@ -235,9 +235,21 @@ def scan_lua_description(path: str, unit: str) -> None:
...
-                Liveries.scan_lua_code(code, path, unit)
+            # Known encodings used for description.lua files
+            encodes = ["utf-8", "ansi"]
+            scanned = False
+            for enc in encodes:
+                with open(description_path, "r", encoding=enc) as file:
+                    try:
+                        code = file.read()
+                    except UnicodeDecodeError:
+                        file.close()
```
DanAlbert, line 246: This is unnecessary, btw. The `with` does this.

Raffson: > This is unnecessary, btw. The `with` does this.

I wasn't sure due to the continue statement, so I played it safe.

Should I remove it again in my other PR?

DanAlbert: It's NBD, just a tip for the future. https://docs.python.org/3/reference/datamodel.html#context-managers explains how `with` works if you're interested.

Raffson: Unfortunately the link you mentioned doesn't really answer my question in this specific case, however I can find numerous examples on stack-overflow where jump-statements are used from within a context manager. So theoretically I conclude that closing the file is redundant indeed, though this particular case gave me the creeps 😂

Anyway, after work I'll probably end up adjusting it 👍

PR #249, opened 2022-08-15
Delay initialization of livery scanner

Raffson:
By replacing the 'Liveries' class-variable with a class-method we can delay the the initialization of the livery scanner.
This is particularly helpful for Liberation since the livery scanner was initializing before Liberation could finish its logging setup.
Thus, none of the logging in the livery scanner would show up in Liberation's logs.

The second commit is a single line to ensure the Liveries-map is initialized before attempting to generate the stub, something I seem to have missed...

### DIFF
--- a/dcs/unittype.py
+++ b/dcs/unittype.py
@@ -1,5 +1,5 @@
 import dcs.lua as lua
-from dcs.liveries_scanner import LiverySet
+from dcs.liveries_scanner import Liveries, LiverySet
 from dcs.payloads import PayloadDirectories
 import re
 import sys
@@ -84,8 +84,7 @@ class FlyingType(UnitType):
     property_defaults: Optional[Dict[str, Any]] = None
 
     pylons: Set[int] = set()
-    livery_name: Optional[str] = None
-    Liveries: LiverySet = LiverySet()
+    livery_name: str = ""  # Aircraft's liveries folder, e.g.: "A-10C II"
     # Dict from payload name to the DCS payload structure. None if not yet initialized.
     payloads: Optional[Dict[str, Dict[str, Any]]] = None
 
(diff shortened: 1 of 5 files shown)


Raffson:
Now that the ANSI support got merged, upon review I realize I forgot to do the same for description files that are read from zip-files. I'm working on a refactor to fix this, so it might be interesting to hold of on the merge until that's ready.

dcs/unittype.py
```diff
@@ -84,8 +84,7 @@ class FlyingType(UnitType):
     property_defaults: Optional[Dict[str, Any]] = None
 
     pylons: Set[int] = set()
-    livery_name: Optional[str] = None
-    Liveries: LiverySet = LiverySet()
+    livery_name: str = ""
```
Raffson, line 87: This change may not be that obvious, but that's cause mypy complains about `livery_name` possibly being `None`, which violates the typing in `Liveries.__getitem__`. But looking back at this I wonder if it might break something w.r.t. mission generation. Any thoughts?

rp-: theoretically it could break, as now it is expected that `livery_name` is always not None.
But I guess you should know best now, if this really has some impact?

Raffson: Afaik an empty string as livery will fall back to DCS' default. The `default_livery` method in unittype.py also returns an empty string in case it can't find one, so I would assume this isn't going to break things.

Despite that, it might not hurt to wait until I can confirm there's nothing that depends on `livery_name` being `None`.

DanAlbert: > an empty string as livery will fall back to DCS' default

This would make a great code comment.

Raffson: > > an empty string as livery will fall back to DCS' default
> 
> 
> 
> This would make a great code comment.

In general, any invalid livery should fall back to tet default. However I remember having a "missing textures" issue when testing the F-15D, so I think there's a caveat to it but I'd have to do some tests to double check.

Raffson: This is gonna sound funny 🤣 

So my last comment is totally irrelevant, fact is I'm the one who added to class variable in the first place, and it's not representing a livery name/id, but rather the name of the aircraft's livery folder, e.g. "A-10C II".

So yeah, this shouldn't be a problem at all. But in order to avoid confusion in the future, I'm gonna add a comment to it.

PR #255, opened 2022-08-24
Add 'is_point_in_zone' for TriggerZones

Raffson:
Also fixed an edge case in 'Polygon.point_in_poly', other than that it's all tests.


dcs/mapping.py
```diff
@@ -296,6 +296,8 @@ def __init__(self, terrain: Terrain, points: List[Point] = None):
 
     def point_in_poly(self, point: Point) -> bool:
         """Checks if the given point is within the polygon.
+        If the polygon represents a line/point,
+        the test checks whether the point is on the line/point.
```
bobmoretti, line 300: This is a weird/probably nonstandard mathematically... typically `point_in_poly()` should return `False` if the point lies on the boundary of the polygon.

As in, the "edge case" you identified was actually correct behavior.

Convention aside, what concrete benefit does this algorithm change afford to a user of PyDCS?

bobmoretti (changes requested):

dcs/mapping.py
Raffson: > This is a weird/probably nonstandard mathematically... typically `point_in_poly()` should return `False` if the point lies on the boundary of the polygon.
> 
> 
> 
> As in, the "edge case" you identified was actually correct behavior.

According to what source if you don't mind me asking?

bobmoretti: This is coming down to overly pedantic details, but if the point is _inside_ a polygon, I expect it to not be on the boundary. Rather, I expect it to be in the interior of the polygon.

For practical purposes, I don't see how it matters in PyDCS, or why it should be changed.

Raffson: Suppose you'd want to connect 2 buildings with a trigger, a quad-point would allow you to connect the buildings using a line. I know it's a special case, and it's easily removed too, but I still fail to see how it's mathematically nonstandard. I don't see how this is any different from asking whether a point is on a line.

PR #258, opened 2022-08-25
Add 'Restrict Targets' waypoint option

bobmoretti:
Added missing "Restrict targets" task option.

Added test cases including to `loadtest.miz` and synthesize a mission from within PyDCS.


bobmoretti:
This conflicts with #257 (loadtest.miz). Once that is merged I'll fix the conflict, update the branch, and convert this to a full PR.

bobmoretti:
Updated the test, sorry for the delay.

Issue #69, opened 2020-08-28
flesh out airfield info


DanAlbert:
A lot of airfield information seems WIP: https://github.com/pydcs/dcs/blob/master/dcs/terrain/caucasus.py (other maps as well)

The specific pieces I'm interested in:

* TACAN
* ILS
* ATC frequencies

rp-:
if it is somewhat accessible via lua, it can be exported, its just pretty cumbersome,
as there is no public API, to get the values from the correct places and takes a lot of DCS restarts until the export script is working

DanAlbert:
I poked around a little bit to see how tractable this was, and unfortunately it doesn't seem that accessible. The data is definitely there in a few other lua files, but I couldn't find where it was exported. The `<DCS>/Mods/terrains/<terrain>/beacons.lua` file has the following:

```
{
    display_name = _('INCIRLIC');
    beaconId = 'airfield16_0';
    type = BEACON_TYPE_VORTAC;
    callsign = 'DAN';
    frequency = 108400000.000000;
    channel = 21;
    position = { 222639.437500, 73.699811, -33216.257813 };
    direction = 0.000000;
    positionGeo = { latitude = 37.015611, longitude = 35.448194 };
    sceneObjects = {'t:124814096'};
};
{
    display_name = _('');
    beaconId = 'airfield16_1';
    type = BEACON_TYPE_ILS_LOCALIZER;
    callsign = 'IDAN';
    frequency = 109300000.000000;
    position = { 222315.812500, 73.638219, -33655.464844 };
    direction = -125.002995;
    positionGeo = { latitude = 37.012552, longitude = 35.443402 };
    sceneObjects = {'t:356059568'};
    chartOffsetX = 3617.000000;
};
{
    display_name = _('');
    beaconId = 'airfield16_2';
    type = BEACON_TYPE_ILS_GLIDESLOPE;
    callsign = 'IDAN';
    frequency = 109300000.000000;
    position = { 220377.484375, 48.478762, -36217.171875 };
    direction = -125.000001;
    positionGeo = { latitude = 36.994247, longitude = 35.415483 };
    sceneObjects = {'t:716015361'};
    chartOffsetX = 3617.000000;
};
{
    display_name = _('');
    beaconId = 'airfield16_3';
    type = BEACON_TYPE_ILS_GLIDESLOPE;
    callsign = 'DANM';
    frequency = 111700000.000000;
    position = { 222001.125000, 63.883437, -34347.218750 };
[...]

PR #265, opened 2022-09-21
Preserve beacon information in Airport and Runway.

DanAlbert:
These mappings alone aren't very useful, but it's also possible to extract the data for each beacon so we can associate these IDs with TACAN and ILS info.

https://github.com/pydcs/dcs/issues/69

### DIFF
--- a/dcs/beacons.py
+++ b/dcs/beacons.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Dict
+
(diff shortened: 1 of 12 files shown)


DanAlbert:
Tested with liberation (which has the data backing those beacon IDs), and I was able to associate TACAN and ILS beacons to runways and airfields using this. Will submit for now and clean up by making `tacan` and `ils` properties of the pydcs types once I decide the best way to work that data in.

PR #269, opened 2022-10-30
Export DCS 2.8

Raffson:
Introduces the MB-339


Raffson:
Commented out everything related to `Ka_50_3` in `countries.py` since there's no such helicopter being exported to `helicopters.py`. I assume this is preliminary stuff from ED for the black shark 3 upgrade.

DanAlbert:
> Commented out everything related to Ka_50_3 in countries.py

Manually? The export script will need to be fixed.

DanAlbert:
(if not manually, just omit those lines during export rather than the commented lines)

Raffson:
> Manually? The export script will need to be fixed.

Adding `Ka_50_3` to `countryHeliIgnore` did the trick.

Raffson:
While at it, should `MB_339A` be removed from `countryPlaneIgnore`?

DanAlbert:
That seems likely. Good catch.

Raffson:
Should be ready 👍

PR #270, opened 2022-11-11
Add option to jettison empty tanks

Raffson:
Noticed this was missing, solution's pretty much the same as any other boolean option

### DIFF
--- a/dcs/action.py
+++ b/dcs/action.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Type
+from typing import Any, Dict, List, Type, Optional
 from dcs.lua.serialize import dumps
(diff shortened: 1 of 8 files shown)


DanAlbert:
It looks like the errors are in files you didn't touch (maybe we don't pin the mypy version?), but because of that I can't know if this has other errors or not. I'll fix the build later and rerun the checks, but this _looks_ right.

Raffson:
> maybe we don't pin the mypy version?

That's probably it

Raffson:
I guess these were errors that weren't being caught previously, but they're definitely legit so I went ahead and quickly fixed it in a second commit

PR #275, opened 2022-12-17
Export

Raffson:
Export for version 2.8.1.34437


Raffson:
Looks like the export script needs another adjustment for outputting property defaults

tools/pydcs_export.lua
```diff
@@ -426,6 +427,8 @@ from dcs.liveries_scanner import Liveries
                         defval = 'True'
                     elseif defval == false then
                         defval = 'False'
+					elseif type(defval) == 'string' then
```
DanAlbert, line 430: Spaces, not tabs

PR #280, opened 2022-12-23
Add support for static rail cars

bobmoretti:
These don't have a `canCargo` or `catgeory` field in the mission table, when saved with the mission editor.

Without the associated changes to `unit.py`, the test cases here fail.

PR #282, opened 2023-01-03
Don't delete effect presets for smoke effects

bobmoretti:
Prior to this change, PyDCS would remove custom smoke effect attributes, effectively resetting all of their sizes to the default effect.

This change adds optional fields that appear for smoke effect in-game static objects.


bobmoretti:
This will conflict once #280 is merged.

DanAlbert (approved):
LGTM, but as you said, conflicts :)

bobmoretti:
@DanAlbert  thanks, conflicts resolved :)

PR #290, opened 2023-02-23
added color to triggers

davidp57:
Hi,

I added the ability to set a color for a trigger.

See this in action here:
image

### DIFF
--- a/dcs/triggers.py
+++ b/dcs/triggers.py
@@ -11,11 +11,12 @@ from dcs import condition
 if TYPE_CHECKING:
     from dcs.terrain import Terrain
 
-class TriggerColor(Enum):
+class TriggerColor(str, Enum):
     Red = "0xff0000ff"
     Green = "0x00ff00ff"
     Blue = "0x0000ffff"
     Yellow = "0x00ffffff"
+    White = "0xffffffff"
 
 class TriggerZoneType(IntEnum):
     Circular = 0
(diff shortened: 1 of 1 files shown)


dcs/triggers.py
```diff
@@ -11,6 +11,11 @@
 if TYPE_CHECKING:
     from dcs.terrain import Terrain
 
+class TriggerColor(Enum):
+    Red = "0xff0000ff"
```
rp-, line 15: Maybe add the `White` also here?
I guess the rest of the colors are the same as in DCS predefined?

dcs/triggers.py
```diff
@@ -197,6 +202,7 @@ class TriggerRule:
 
     def __init__(self, event: Event, comment: str):
         self.comment: str = comment
+        self.color: str = "0x00000000"
```
rp-: and then also use the `TriggerColor.White` here

rp- (changes requested):

dcs/triggers.py
davidp57: Hello,
This has been done in 18eb279!
Thanks.

rp-: Seems I looked at the wrong diff, sorry

rp-:
Just fix the lint errors and we are good to go

davidp57:
done (I think)

PR #294, opened 2023-04-14
Fixed issue with airplanes trying to spawn on heli slots

Lordeath19:
Some airfields such as Tbilisi in Caucuses and Aleppo in Syria have Heli only slots, these slots are currently being occupied by normal airplanes due to `terrain.py` not taking Heli only slots into account. This causes planes to start in air and crash immediately at mission start.

This fix adds the check and some tests to make sure airplanes don't spawn in Heli only slots


DanAlbert:
Now if only the `shelter` and `large` properties were actually used... Thanks.

PR #297, opened 2023-05-09
Factor out wingreg usage to simplify.

DanAlbert:
Querying the registry in each caller made the test mocks quite messy. Factor that out so we can clean up the implementation and the tests. This also removes the "is standalone" and "is steam" APIs since those don't seem useful and they complicated the implementation. The other cleanup this makes easy is switching to the contextmanager OpenKey, which fixes the bug I found when writing the test in the previous PR.


DanAlbert:
Most of the code here wasn't previously type checked because it declared no types, and that's why it's failing. I'll need to fix that too.

PR #316, opened 2023-05-19
Livery parser fix.

DanAlbert:
Make the livery parser tolerant of its own failures. Our lua parser is not robust enough to have failures be fatal since DCS can arbitrarily push an update that will find our bugs. I've added an xfail test for the new failure case, but haven't implemented the fix in the parser.


DanAlbert:
Text got pulled out of the PR. The github action doesn't like pytest xfail for some reason

PR #323, opened 2023-07-03
Prevent liveries with invalid country values from stopping livery processing.

Malakhit:
This should hopefully fix the issues I've been having with DCS Liberation new game creation.

### DIFF
--- a/dcs/liveries/livery.py
+++ b/dcs/liveries/livery.py
@@ -90,7 +90,11 @@ class Livery:
         if countries_table is None:
             countries = None
         else:
-            countries = set(countries_table.values())
+            try:
+                countries = set(countries_table.values())
+            except AttributeError:
+                logging.exception("Could not get country information from livery at %s", path)
+                countries = None
         order = data.get("order", 0)
 
         order = None if path_id == "default" else order


Malakhit:
image

I probably have a dodgy livery somewhere - with over 1000 liveries installed, I wouldn't be surprised!

DanAlbert:
Would it be better to improve the error message so you can uninstall the busted livery?

DanAlbert:
That is, improve, but still fail. Doesn't seem helpful to ignore (no one reads logs) broken data.

Malakhit:
That's the thing - the livery probably isn't busted from a functionality point-of-view, it's just got a countries block in it that PyDCS can't process. I've not come across a livery that the game refuses to load at this point - it might have some missing textures but that won't cause DCS any issues.

DanAlbert:
If the livery isn't broken then this is definitely the wrong fix. What other format of the countries field needs to be handled?

Malakhit:
I'm not sure which livery is causing the issue on my end, but at a guess, it's either going to be:
- a livery file with a countries block that spans multiple lines
- a livery file with a commented out countries block
- a livery file with a countries block with no entries

I generally try to tidy up livery description files where possible, but due to the volume of them it's a losing battle. It would be useful if it _could_ log which files need modifying, but it would also be beneficial if a livery with a dodgy countries block wouldn't cause an unhandled exception upon trying to start a new Liberation campaign. I figured that just having a default of "None" for countries, and only populating the field if it can be parsed from the lua file, would make sense.

DanAlbert:
> I figured that just having a default of "None" for countries

`None` isn't a default, it's a value that means "any country". Without you narrowing down which livery has the behavior difference there's no way of knowing if this change introduces a bug (pydcs reports a livery valid for any country, but DCS rejects it) or not.

You've got a pretty easy opportunity to diagnose the real bug here, right? (use the log message in this PR to find the livery with the different field format, look at it). If that's easy to fix in pydcs, may as well just fix it. If it's not easy to fix, then we should ignore the livery entirely rather than load bad data. Seems silly to go straight for that without at least trying to fix the problem though.

In any case, you should add a test (this corner of pydcs has 100% coverage or close to it), and to do that you'll need to figure out the offending format.

Malakhit:
Okay, well I've now fixed the liveries that it detected which were incorrectly formatted. Folks on user files have a variety of interpretations of what is and isn't valid lua, it would seem. Closing this as it's no longer needed.

DanAlbert:
I am curious what the bad content was, if you're able to share. You're right that pydcs shouldn't crash on broken mod content, I just want to make sure we have a test covering the bad case.

Malakhit:
> I am curious what the bad content was, if you're able to share. You're right that pydcs shouldn't crash on broken mod content, I just want to make sure we have a test covering the bad case.
> 
> But I'm not too concerned about it (garbage in, garbage out, after all), so if you're happy with it as is we don't need to spend more time on it :)

The bad content was malformed lua. Missing closing braces, interesting attempts at making lists, closing braces with more definitions outside of them, and so on.

PR #338, opened 2023-11-17
Put SetUnlimitedFuelCommand in wrappedactions dict

magwo:
Doesn't seem to work otherwise.


magwo:
I think this was missing in your SetUnlimitedFuel commit @DanAlbert  :)

DanAlbert:
Interesting, because it showed up correctly in the ME. I would be this is needed for deserializing rather than serializing?

magwo:
Yes, deserializing.

PR #346, opened 2023-12-11
Loads and saves requiredModules

332fg-raven:
Loads and saves requiredModules section.


dcs/mission.py
```diff
@@ -260,6 +260,9 @@ def loaddict(fname: str, mizfile: zipfile.ZipFile, reserved_files: List[str]) ->
 
         imp_mission = mission_dict["mission"]
 
+        # required modules
+        self.requiredModules = imp_mission.get("requiredModules", None)
```
DanAlbert, line 264: nit: `snake_case`

tests/test_mission.py
```diff
@@ -872,3 +872,15 @@ def preset_preserved(group: dcs.unitgroup.StaticGroup,
 
         static_groups = m2.coalition["blue"].countries["USA"].static_group
         self.assertTrue(any(preset_preserved(g, 3) for g in static_groups))
+
+    def test_smoke_required_modules(self) -> None:
+        mizname = "missions/Mission_with_required_modules.miz"
+        m = dcs.mission.Mission()
+        m.load_file(mizname)
+        self.assertTrue(len(m.requiredModules.keys()) == 1)
```
DanAlbert, line 880: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual

Better to assert that the contents are actually preserved :)

DanAlbert (changes requested):
Minor review notes, but since the checks didn't pass anyways, might as well fix them at the same time.

Otherwise LGTM, and thanks!

DanAlbert:
btw for the other problem I see there's also an old and overlooked PR https://github.com/pydcs/dcs/pull/219 that aims to do similarly, but with a different approach (that I haven't looked at so idk if it has the same problems or not).

dcs/mission.py
332fg-raven: n/p, thought to keep it in sync with what's in mission file.

tests/test_mission.py
332fg-raven: It's a dictionary, not a list, hence assertEqual

DanAlbert: The exact type wasn't my point though. Assert the contents, not the size. You've still got `assertTrue(something about the size)` where instead of `assertEqual(expected, the contents)`

PR #347, opened 2023-12-19
Reupload of #344

DanAlbert:
I goofed when force pushing https://github.com/pydcs/dcs/pull/344 and pushed pydcs/master to that branch instead of my local HEAD. Doing that automatically closed the PR since it no longer had any new commits in it, and because there's no longer a PR from that branch I can't force push to it any more. Reuploading.


gregretkowski:
thanks for merging this!

PR #348, opened 2023-12-20
Default removal of Apache FCR to False now that FCR has been released

zhexu14:
Now that the FCR has been released, default the removal of the FCR to false.

### DIFF
--- a/dcs/helicopters.py
+++ b/dcs/helicopters.py
@@ -1081,7 +1081,7 @@ class AH_64D_BLK_II(HelicopterType):
     }
 
     property_defaults: Dict[str, Any] = {
-        "FCR_RFI_removed": True,
+        "FCR_RFI_removed": False,
         "PltNVG": True,
         "CpgNVG": True,
         "FlareBurstCount": 0,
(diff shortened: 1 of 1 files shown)


DanAlbert:
These are auto-generated by the export script, which I've done in https://github.com/pydcs/dcs/pull/350.

These properties haven't changed in DCS. The defaults here aren't actually our preference, they're the default value that DCS uses when you create a new aircraft in the mission editor. DCS apparently still defaults Apaches to not including the FCR, so no change required.

But since I know you're a fellow Liberation dev, I assume you're trying to change Liberation's default? We've got a separate system for when we want to override the DCS defaults with our preferences. Look at `default_overrides` in the unit yamls: https://github.com/dcs-liberation/dcs_liberation/blob/1efce862fb2cc3330a3cfaae372ffd6aefbf63a9/resources/units/aircraft/F-16C_50.yaml#L53

zhexu14:
Awesome, thanks for pointing me in the right direction, that was exactly what I was trying to do. I'll open up a PR in the Liberation repo.

PR #349, opened 2023-12-20
Feature/picture to all wwii follow big formation carpet bombing and opt min alt restriction

332fg-raven:
Added features required to parse the WWII mission file included in the commit:
- a_out_picture action
- FollowBigFormation
- CarpetBombing
- option Min Alt Restriction

If you come up with a better than for WWIIFollowBigFormationOrder, feel free to suggest.

Some values (such as weapon type) are hard-coded, but will be added separately.


dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
         return d
 
 
+class PictureToAll(Action):
+    predicate = "a_out_picture"
+
+    def __init__(self, file_res_key: Optional[ResourceKey] = None):
+        super(PictureToAll, self).__init__(PictureToAll.predicate)
```
DanAlbert, line 1794: For new code, just use `super()`. We don't care about Python 2 compat.

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+        if file_res_key:
```
DanAlbert, line 1795: `if file_res_key is not None` (there's a behavior difference between `None` and a falsy `ResourceKey`)

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+            self.file_res_key = file_res_key
```
DanAlbert, line 1796: This needs to happen outside the `if`. As written, this object has different members depending on the arguments passed.

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+            self.params.append(self.file_res_key)
+
+    @classmethod
+    def create_from_dict(cls, d, mission):
+        return cls(ResourceKey(d["file"]))
```
DanAlbert, line 1801: Based on how you've written the rest of the class, this looks wrong. Is the `ResourceKey` optional or not?

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
         return d
 
 
+class PictureToAll(Action):
+    predicate = "a_out_picture"
+
+    def __init__(self, file_res_key: Optional[ResourceKey] = None):
```
DanAlbert, line 1793: Annotate the return type too. `-> None`

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+        super(PictureToAll, self).__init__(PictureToAll.predicate)
+        if file_res_key:
+            self.file_res_key = file_res_key
+            self.params.append(self.file_res_key)
+
+    @classmethod
+    def create_from_dict(cls, d, mission):
```
DanAlbert, line 1800: Type annotations

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+        return cls(ResourceKey(d["file"]))
+
+    def dict(self):
```
DanAlbert, line 1803: Type annotations

dcs/action.py
```diff
@@ -1787,6 +1787,25 @@ def dict(self):
...
+        d = super(PictureToAll, self).dict()
+        d["file"] = self.file_res_key.key
```
DanAlbert, line 1805: This is wrong. `self.file_res_key` won't be defined by the default constructor. Once you've fixed that, it still might be `None`.

dcs/action.py
```diff
@@ -1865,5 +1884,6 @@ def dict(self):
     "a_set_ai_task": AITaskSet,
     "a_remove_scene_objects": RemoveSceneObjects,
     "a_scenery_destruction_zone": SceneryDestructionZone,
-    "a_zone_increment_resize": ZoneIncrementResize
+    "a_zone_increment_resize": ZoneIncrementResize,
+    "a_out_picture": PictureToAll
```
DanAlbert, line 1888: Trailing comma (I know there wasn't one before, but there should have been)

dcs/mapping.py
```diff
@@ -97,6 +97,11 @@ class Vector2:
     x: float
     y: float
 
+@dataclass
```
DanAlbert, line 100: Unless there's a good reason not to, `frozen=True`

dcs/mapping.py
```diff
@@ -97,6 +97,11 @@ class Vector2:
...
+class WWIIFollowBigFormationOrder:
```
DanAlbert, line 101: This should probably go in `task.py`. It's got nothing to do with mapping afaict.

dcs/mapping.py
```diff
@@ -97,6 +97,11 @@ class Vector2:
...
+    x: int # distance
```
DanAlbert, line 102: The comments are better names. DCS might call them x/y/z, but it's a damned confusing API. The pydcs API doesn't need to exactly match the miz structure.

Include units in the name when you rename them. `distance_meters` or whatever this happens to be.

`int` is also incorrect. Lua doesn't have integers, only floats.

dcs/mapping.py
```diff
@@ -97,6 +97,11 @@ class Vector2:
...
+    y: int # elevation
```
DanAlbert, line 103: And on this one, `_meters_msl` (or `_ft_agl`, or whatever it happens to be)

dcs/mapping.py
```diff
@@ -97,6 +97,11 @@ class Vector2:
...
+    z: int # interval
```
DanAlbert, line 104: What is this really? Distance between individual aircraft? Distance between subgroups? I don't know how big formation works so it's not obvious to me. Whichever it is, use a name that makes it clear.

dcs/action.py
DanAlbert, line 1793: Is the file key actually optional? That's up to DCS and DCS can sometimes be surprising, but that does surprise me.

dcs/task.py
```diff
@@ -1026,6 +1026,80 @@ def __init__(self, from_index=None, to_index=None):
...
+    :param last_wpt_index: ???
+    :param formation_type: ???
+    :param pos_in_box: ???
+    :param pos_in_group: ???
+    :param pos_in_wing: ???
+    """
+    Id = "FollowBigFormation"
+
+    def __init__(self,
+                 orders: WWIIFollowBigFormationOrder = WWIIFollowBigFormationOrder(0, 0, 0),
```
DanAlbert, line 1047: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

dcs/task.py
```diff
@@ -1026,6 +1026,80 @@ def __init__(self, from_index=None, to_index=None):
...
             self.params["nWaypointIndx"] = to_index
 
+class WWIIFollowBigFormation(Task):
+    """Follow as big formation task
+
+    If added to a group, the group will follow the big formation according to orders
+
+    :param group_id: ID of a group to follow
+    :param orders: :py:class:`dcs.mapping.WWIIFollowBigFormationOrder` how to follow
+    :param last_wpt_index_flag: ??? -
```
DanAlbert, line 1036: If you're going to document them (and thanks for doing that!), document them :)

dcs/task.py
```diff
@@ -1026,6 +1026,80 @@ def __init__(self, from_index=None, to_index=None):
...
+    :param formation_type: ???
+    :param pos_in_box: ???
+    :param pos_in_group: ???
+    :param pos_in_wing: ???
+    """
+    Id = "FollowBigFormation"
+
+    def __init__(self,
+                 orders: WWIIFollowBigFormationOrder = WWIIFollowBigFormationOrder(0, 0, 0),
+                 group_id: Optional[int | None] = None,
```
DanAlbert, line 1048: Really? What's the behavior when this is `None`?

dcs/task.py
```diff
@@ -1026,6 +1026,80 @@ def __init__(self, from_index=None, to_index=None):
...
+                 last_wpt_index_flag: Optional[bool] = True,
```
DanAlbert, line 1049: Are these really tri-states? They don't look like tri-states. How does `True` differ in behavior from `None`?

dcs/task.py
```diff
@@ -1026,6 +1026,80 @@ def __init__(self, from_index=None, to_index=None):
...
+            "pos": {
+                "x": orders.x,
+                "y": orders.y,
+                "z": orders.z
+            },
+            "groupId": group_id,
+            "lastWptIndexFlag": last_wpt_index_flag,
+            "lastWptIndexFlagChangedManually": last_wpt_index_flag_changed_manually,
+            "lastWptIndex": last_wpt_index,
+            "formationType": formation_type, # 0 = 1943 style Combat Box, 1 = 1942 JAVELIN DOWN
```
DanAlbert, line 1065: Add enums for these.

DanAlbert (changes requested):
Please upload this as mutliple PRs. I have so many comments so far that it's going to be very difficult to get this merged as a single PR. I stopped reviewing around my last comment, so there's more review to be done, but it's only going to be harder to track the review comments post-split if I keep going.

332fg-raven:
I understand. Can you give me a good starting point on the code that follows the rules? For PictureToAll I used SoundToAll as a starting point, for WWIIFollowBigFormation I used Land. About documenting fields in DCS I wish I could, but I have no idea what they mean - I was looking at MissionEditor lua files to try to understand things, but then again, I think it's safer to just remove the comment. I'll clean that up, no worries. I can create multiple PR's, but you're asking a lot before xmas :-)

DanAlbert:
The checks that are failing cover our main rules. You need (correct) type annotations, and must pass a subset of flake8 rules (which you can find by looking at the github action definition in the .github directory).

The bulk of my comments aren't about conventions though, they're about correctness and API design. Primarily two big things:

1. We must be able to read and write DCS miz files correctly, so at the `dict()` and `create_from_dict()` methods are often pretty wonky, but they should never be wonkier than DCS itself. For example, if DCS allows you to create a big formation task without a group to follow, those methods should support deserializing those tasks, and need to re-serialize something with identical behavior (even if it's not the identical table). If they _don't_ allow that, we shouldn't either.
2. We should expose useful APIs to pydcs consumers, which _sometimes_ are a direct match for the DCS miz structure, but not usually. The DCS miz names are usually unhelpful (I bet ED even agrees, but they can't rename them without breaking old miz files, and they're an implementation detail anyway), so we want to give good names in our APIs. Magic numbers and strings require users to go digging through the source to learn how to use the API, so we should add enums for those things instead.

   The part of this where there is no strict rule and it requires some judgement is whether we ought to allow callers to pass `None` for things that the mission editor doesn't strictly reject but also make no sense. Is there any situation where someone would _intentionally_ create a big formation task that flies formation with nobody? Probably not, so we might as well disallow that in the type system to help people catch bugs. (I'm wrong on this specific example, see below)

I just checked that specific example (can a big formation follow nobody?), and apparently yes.

image

That means that we do need to be able to serialize and deserialize a `None` group ID for that task (assuming that's represented with `nil` in the miz and not the absence of the property or `0` or `""` or something else, you have to look at the miz to be sure). Looking at how the task works in the ME, I'm guessing it's also _expected_ that one group in the formation (the leader, thought they don't necessarily need to be in the lead position of the formation) uses "nothing", so it's probably correct (and required) to support `None` as an argument in our API as well.

So that part of the big formation API is correct, but pretty non-obvious IMO so worth adding some comments.

The rest of the comments were typical python correctness stuff, not anything specific to pydcs.

One thing that might be causing confusion: do you understand what `Optional[T]` actually does? It doesn't mean that the argument is optional because there's a default argument. it means that `None` is a valid value. `Optional[int]` is "some integer or `None`", not "some integer or don't pass any argument at all for the default".

> About documenting fields in DCS I wish I could, but I have no idea what they mean - I was looking at MissionEditor lua files to try to understand things

The DCS manual is usually pretty good about that, actually. Not always, but these are pretty old tasks and IME the older stuff is usually better documented. If it's not documented... it might be better to just not expose those arguments. If we don't know what they do it's pretty hard to design the right API for them.

> but you're asking a lot before xmas :-)

Fortunately there are no deadlines here :)

332fg-raven:
No issues, opened up a new PR which adds support for PictureActions. Closing this PR and will address the rest separately.

PR #351, opened 2023-12-21
Added support for a_out_picture action

332fg-raven:
* supports sending picture to all, coalition, country, group and unit. DCS UI issue allows picture to be sent to unit without picture resource, so a dedicated test covers that.

NOTE: assert isinstance is used to make mypy happy. If there are better ways, please let me know.


dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
         return d
 
 
+class PictureAction(Action):
+    def __init__(self, predicate: str, file_res_key: ResourceKey, seconds: int,
+                 clearview: bool, start_delay: int,
+                 horz_alignment: str, vert_alignment: str,
```
DanAlbert, line 1793: Enums for both of these.

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+                 size: int, size_units: str) -> None:
```
DanAlbert, line 1794: And for `size_units`

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+        super().__init__(predicate)
+        self.file_res_key = file_res_key
+        self.seconds = seconds
+        self.clearview = clearview
+        self.start_delay = start_delay
+        self.horz_alignment = horz_alignment
+        self.vert_alignment = vert_alignment
+        self.size_units = size_units
+        self.size = size
```
DanAlbert, line 1803:
```
if size not in range(1, 101):
    raise ValueError
```

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+                   d["horzAlignment"], d["vertAlignment"], d["size"], d["size_units"])
+
+    def dict(self):
+        return super().dict()
+
+
+class PictureToCoalition(PictureAction):
+    predicate = "a_out_picture_s"
+
+    def __init__(self, coalition: str, file_res_key: ResourceKey, seconds: int,
```
DanAlbert, line 1849: The type here should be `Coalition`

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+    def dict(self):
+        d = super().dict()
+        d["coalitionlist"] = self.coalition
+        return d
+
+
+class PictureToCountry(PictureAction):
+    predicate = "a_out_picture_c"
+
+    def __init__(self, country: int, file_res_key: ResourceKey, seconds: int,
```
DanAlbert, line 1875: Two options here:

1. Use `Country` instead of `int`
2. Rename it `country_id`

1 is a bit nicer on the caller, but 2 might be better for testing? Up to you.

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+    def dict(self):
+        d = super().dict()
+        d["countrylist"] = self.country
+        return d
+
+
+class PictureToGroup(PictureAction):
+    predicate = "a_out_picture_g"
+
+    def __init__(self, group: int, file_res_key: ResourceKey, seconds: int,
```
DanAlbert, line 1901: Same here as for country. Can `Group` be used instead of `int`? If not, `group_id: int`.

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+    def dict(self):
+        d = super().dict()
+        d["group"] = self.group
+        return d
+
+
+class PictureToUnit(PictureAction):
+    predicate = "a_out_picture_u"
+
+    def __init__(self, unit: int, file_res_key: ResourceKey, seconds: int,
```
DanAlbert, line 1927: Same as the previous two, but for `unit`

DanAlbert (changes requested):

DanAlbert:
> NOTE: assert isinstance is used to make mypy happy. If there are better ways, please let me know.

`isinstance` is correct here.

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+        self.file_res_key = file_res_key
+        self.seconds = seconds
+        self.clearview = clearview
+        self.start_delay = start_delay
+        self.horz_alignment = horz_alignment
+        self.vert_alignment = vert_alignment
+        self.size_units = size_units
+        self.size = size
+
+    def dict(self) -> Dict[Any, Any]:
```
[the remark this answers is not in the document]
332fg-raven: Yes.

dcs/action.py
```diff
@@ -1787,6 +1787,166 @@ def dict(self):
...
+        d["seconds"] = self.seconds
+        d["clearview"] = self.clearview
+        d["start_delay"] = self.start_delay
+        d["horzAlignment"] = self.horz_alignment
+        d["vertAlignment"] = self.vert_alignment
+        d["size_units"] = self.size_units
+        d["size"] = self.size
+        return d
+
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
332fg-raven: ok.

dcs/action.py
332fg-raven: Same here, changed to group_id.

dcs/action.py
332fg-raven: This is now unit_id -- if you can help me solve circular import, I don't mind changing it to proper types.

dcs/action.py
```diff
@@ -1787,6 +1787,195 @@ def dict(self):
...
 
 
+class PictureAction(Action):
+
+    class HorzAlignment(Enum):
+        Left = "0"
+        Center = "1"
+        Right = "2"
+
+        def __eq__(self, other) -> bool:
```
DanAlbert, line 1797: I'm surprised to see `__eq__` defined for these. Are these actually needed?

If they are, `other: Any` (on all of them)

dcs/action.py
```diff
@@ -1787,6 +1787,195 @@ def dict(self):
...
+        Center = "1"
+        Bottom = "2"
+
+        def __eq__(self, other) -> bool:
+            if isinstance(other, str):
+                return self.value == other
+            return self == other
+
+    class SizeUnits(Enum):
+        Original_Size = "0"
```
DanAlbert, line 1813: `OriginalSize` or `ORIGINAL_SIZE` (PEP8 prefers the latter, we have a mix, I suppose lean toward PEP8)

dcs/action.py
```diff
@@ -1825,45 +1857,42 @@ class PictureToAll(PictureAction):
...
                          start_delay=start_delay, horz_alignment=horz_alignment,
-                         vert_alignment=horz_alignment, size=size, size_units=size_units)
+                         vert_alignment=vert_alignment, size=size, size_units=size_units)
         self.coalition = coalition
         self.params = [self.coalition, self.file_res_key, seconds, clearview,
                        start_delay, horz_alignment, vert_alignment, size, size_units]
 
     @classmethod
-    def create_from_dict(cls, d, mission):
+    def create_from_dict(cls, d, mission) -> "PictureToCoalition":
```
DanAlbert, line 1891: Need type annotations. There are a couple others missing later in the file too.

DanAlbert, line 1891: Non blocking, but a useful tip you may want to know for the future: you can `from __future__ import annotations` at the top of the file so you can use `PictureToCoalition` here instead of `"PictureToCoalition"` (it doesn't really matter which form you use, but if you prefer to not use strings for "types", now you know how)

dcs/action.py
DanAlbert: Yeah, Python's a bit of an asshole about this. Up at the top of the file you should do:

```python
from __future__ import annotations

# rest of the normal imports
from typing import TYPE_CHECKING


if TYPE_CHECKING:
    from .country import Country
```

It's a bit funky but that is what Python expects us to do for type-only circular imports.

dcs/action.py
332fg-raven: Noted.

dcs/action.py
332fg-raven: Noted.

332fg-raven: Thanks, will try and avoid using strings. If it's too much a hassle for me, I'd prefer pushing this forward then refactoring just that piece.

dcs/coalition.py
```diff
@@ -289,6 +289,14 @@ def find_group(self, group_name, search="exact"):
 
         return None
 
+    def find_group_by_id(self, group_id: int) -> Optional[unitgroup.Group]:
+        for c in self.countries:
+            g = self.countries[c].find_group_by_id(group_id)
+            if g:
```
DanAlbert, line 295: `is not None`

dcs/action.py
```diff
@@ -1901,52 +1909,55 @@ def dict(self) -> Dict[str, Any]:
...
+        self.group = group
+        self.params = [self.group.id, self.file_res_key, seconds, clearview,
                        start_delay, horz_alignment, vert_alignment, size, size_units]
 
     @classmethod
-    def create_from_dict(cls, d, mission) -> "PictureToGroup":
-        return cls(d["group"], ResourceKey(d["file"]), d["seconds"], d["clearview"], d["start_delay"],
+    def create_from_dict(cls, d: Dict[str, Any], mission: Mission) -> PictureToGroup:
+        group = mission.find_group_by_id(d["group"])
+        assert group is not None
```
DanAlbert, line 1954: As in the other PR, don't use `assert` for error handling. Raise an exception.

dcs/country.py
```diff
@@ -81,6 +81,20 @@ def find_group(self, group_name, search="exact"):
...
         return None
 
+    def find_group_by_id(self, group_id: int) -> Optional[Group]:
+        groups = [self.vehicle_group,
+                  self.ship_group,
+                  self.plane_group,
+                  self.helicopter_group,
+                  self.static_group]
+        for search_group in groups:
+            assert isinstance(search_group, list)
```
DanAlbert, line 91: This looks unnecessary, and would prevent the type checker from doing its job if the types ever changed.

dcs/mission.py
```diff
@@ -1808,6 +1808,21 @@ def find_group(self, group_name, search="exact") -> Optional[unitgroup.Group]:
...
+
+        Args:
+            group_id: group identifier assigned by the mission file
+
+        Returns:
+            Group: the group found, otherwise None
+        """
+        for k in self.coalition:
+            g = self.coalition[k].find_group_by_id(group_id)
+            if g:
```
DanAlbert, line 1822: `is not None`

DanAlbert (changes requested):

dcs/action.py
DanAlbert: Yeah, SGTM. Like I said, not blocking.

dcs/action.py
332fg-raven: @DanAlbert - unit_id to Unit can't be done without refactoring the code which parses triggers. At the moment trigger/action parsing code is unaware of the context of coalition/country it's parsing for, therefore unit_id, since it's not unique, cannot be resolved simply. I'd rather keep this as is and refactor the code later that focuses on adding context to trigger/action parsers, then refactor this piece to use it. Thoughts?

dcs/mission.py
332fg-raven: @DanAlbert - g is an object, it cannot be anything else but Group or None. Do you want to be that explicit?

dcs/country.py
332fg-raven: @DanAlbert  - mypy is not happy if it's not there. You wanna argue with it? :-)

image

dcs/mission.py
DanAlbert: But that doesn't mean it can't be false. `__bool__` is a thing.

dcs/country.py
DanAlbert: I kinda do. I'll pull your branch and have a look, because it doesn't make sense to me. Wouldn't be the first time there's a bug in mypy but I don't mind looking.

dcs/action.py
332fg-raven: OK.

dcs/country.py
DanAlbert: I think it has to do with mypy's type inference. If I explicitly type `groups` as `list[list[Group]]` I get:

```
dcs\country.py:86: error: List item 0 has incompatible type "List[VehicleGroup]"; expected "Group[Any, Any]"  [list-item]
dcs\country.py:87: error: List item 1 has incompatible type "List[ShipGroup]"; expected "Group[Any, Any]"  [list-item]
dcs\country.py:88: error: List item 2 has incompatible type "List[PlaneGroup]"; expected "Group[Any, Any]"  [list-item]
dcs\country.py:89: error: List item 3 has incompatible type "List[HelicopterGroup]"; expected "Group[Any, Any]"  [list-item]
dcs\country.py:90: error: List item 4 has incompatible type "List[StaticGroup]"; expected "Group[Any, Any]"  [list-item]
dcs\country.py:92: error: "Group[Any, Any]" has no attribute "__iter__"; maybe "__str__"? (not iterable)  [attr-defined]
```

If I type it as `list[Sequence[Group]]` it works fine, and unlike the assert, the type checker will still prove it correct and emit an error if the type ever changes. That's `from typing import Sequence`, btw :) (or `from collections.abc`... but I think we still support python versions old enough that we need to use the deprecated type from `typing`?)

332fg-raven: @DanAlbert  - Resolved by adding a type hint to groups.

dcs/mission.py
332fg-raven: Added is not None.

dcs/action.py
DanAlbert: Are these needed then? Could you explain why? It looks superfluous.

dcs/country.py
```diff
@@ -82,13 +82,12 @@ def find_group(self, group_name, search="exact"):
         return None
 
     def find_group_by_id(self, group_id: int) -> Optional[Group]:
-        groups = [self.vehicle_group,
-                  self.ship_group,
-                  self.plane_group,
-                  self.helicopter_group,
-                  self.static_group]
+        groups: List = [self.vehicle_group,
```
DanAlbert, line 85: This is equivalent to `List[Any]`, which you definitely don't want (try `print(group.fsdafdsafg)` with this, the type checker will not complain). You need to do what I said in the comment: `List[Sequence[Group]]`.

dcs/action.py
332fg-raven: @DanAlbert, I'd like to use strong types in tests. I can remove this, but is there a way we'd be able to use strong types in tests like: self.assertEqual(m_action.horz_alignment, PictureAction.HorzAlignment.Left) -- looking it from a point of view that I load a mission in Mission Editor, see the names, like Left, and I open a test and I say "aha! understood". If I see "1" in test I have to open .miz and then say "aha! understood. Let me know what you want me to do.

DanAlbert: Ah, I get it now. You want https://docs.python.org/3/library/enum.html#enum.StrEnum.

DanAlbert:
Other than switching to `StrEnum`, looks good now.

332fg-raven:
@DanAlbert - as I read StrEnum is available in python11. Tests are running using 3.8, 3.9, 3.10, 3.11. I'm using python 3.10 and StrEnum is not available. Should I leave it until 3.11 is required or what's do you suggest?

DanAlbert (approved):

DanAlbert:
> @DanAlbert - as I read StrEnum is available in python11. Tests are running using 3.8, 3.9, 3.10, 3.11. I'm using python 3.10 and StrEnum is not available. Should I leave it until 3.11 is required or what's do you suggest?

Ah, good catch. I didn't realize that was so new. Yeah, keep with what you've got.

PR #355, opened 2024-01-10
Coalition countries saved even when no units on the map

332fg-raven:
While I was working with empty mission I noticed after loading/saving with pydcs I couldn't add units on the map anymore and realized that coalitions had no countries assigned.

This patch saves countries belonging to coalition even if those countries have no units on the map.
See the diff with white spaces ignored - I have done minimal changes to keep the code in load_from_dict() intact.


DanAlbert (changes requested):
I've been staring at this for like 20 minutes and I'm still not sure what the actual difference is in the miz. I see you're passing the complete list of countries from the coalition definition to `Coalition.load_from_dict`, but as best as I can tell, if the country wasn't in `d["country"]` anyway, you're not doing anything else with that data. What am I missing?

The one definite behavior change I do see is the key name fix from `neutral` to `neutrals`... was that actually the fix and the rest is not needed?

332fg-raven:
@DanAlbert - the idea of this patch is to save the countries in coalitions even if they don't have any units on the map. I was looking at what I could do and implemented this in what I thought was the least intrusive way. If you create an empty mission, assign countries to coalitions and save it in DCS Mission Editor, then load & save it in pydcs, re-open it in Mission Editor you'll see that coalitions will be empty. If you want to address it any other way, let me know.

dcs/coalition.py
```diff
@@ -89,172 +89,188 @@ def get_name(mission: "Mission", name: str) -> str:
...
-                        imp_unit = sgroup["units"][imp_unit_idx]
-                        static: Static
-                        if imp_unit["type"] == "FARP":
-                            static = FARP(
+
+        # add countries to coalition even if they have no units
+        for idx, country_id in countries_in_coalition.items():
+            _country = countries.get_by_id(country_id)
+            if country_has_units(country_id, d["country"]):
+                country_idx = get_country_idx(country_id, d["country"])
```
DanAlbert, line 111: You've got a `has` and `get` that perform identical operations with the only difference being the returned value. You can rewrite by replacing `country_has_units` and `get_country_idx`

```python
def country_dict_from_id(country_id: int, d: Dict[int, Any]) -> Optional[dict[str, Any]]:
    # Same implementation as get_country_idx, more or less
```

(no idea if that name is correct or not, pick whatever makes sense, that's just an example)

And then rewrite the loop as:

```python
for country_id in countries_in_coalition.values():
    self.add_country(self.countries.get_by_id(country_id)
    imp_country = country_dict_from_id(country_id, d["country"])
    if imp_country is None:
        continue

    # rest of the function as before, though the last line moved up so it wouldn't be skipped
```

Less code and runs faster (it doesn't search for the same value 3 times). (It'd be even easier if the miz table was keyed by country ID rather than by index, but whatever, can't fix that)

Alternatively (and probably better?), put this `for` back the way it was, but move the `self.add_country()` call out into a separate

```python
for country_id in countries_in_coalition.values():
    self.add_country(country_id)
```

outside this loop and then you can remove the lookup functions entirely.

DanAlbert:
I see the line I missed. If I tell the diff to ignore the whitespace, it hides the part that matters from me. If I tell it not to, the PR is unreadable. There's a final line in this gigantic function of `self.add_country(_country)`, and after the PR that is run, whereas before it was not.

Now that I've figured out what's going on, left a comment for a faster and smaller implementation.

dcs/coalition.py
332fg-raven: @DanAlbert - I like the idea of keeping the old code for what it did, then adding at the end of the function a simple loop loop that checks if a country is in the coalitions, skip it (since the orig code processed it), otherwise add it.

332fg-raven:
@DanAlbert - reimplemented using simpler approach, as suggested.

DanAlbert:
Conflicts need to be resolved, but otherwise LGTM.

332fg-raven:
@DanAlbert - rebased to master. Do you prefer rebases or merges from master -> feature to resolve conflicts?

DanAlbert:
> @DanAlbert - rebased to master. Do you prefer rebases or merges from master -> feature to resolve conflicts?

Unless you're carefully maintaining a clean git history in the PR (squashing and fixing-up as needed for review comments), it doesn't make a difference since I'm gonna squash merge anyway.

And unfortunately, needs to be done one more time since it also conflicts with the other PR. (I really wish GitHub had a better way of doing this, but the only ways I know of are third-party add ons that aren't really ready yet.)

332fg-raven:
@DanAlbert - rebased. mypy will fail - review #356 first, which addresses the duplicate tests too, then this can go. I'll do the necessary rebases, no worries.

dcs/coalition.py
```diff
@@ -257,6 +257,13 @@ def load_from_dict(self, mission, d) -> List[StatusMessage]:
                         static_group.add_unit(static)
                     _country.add_static_group(static_group)
             self.add_country(_country)
+
+        # iterate over all .miz countries in coalition, even without any units
+        # on the map, and add them to the respective coalition
+        for country_id in countries_in_coalition.values():
+            if self.country_by_id(country_id) is None:
```
DanAlbert, line 264: This lookup is wasteful. If you delete L259 you don't need it, right? (that was what I'd tried to suggest in my earlier comment, sorry for not being more clear)

DanAlbert:
> @DanAlbert - rebased. mypy will fail - review #356 first, which addresses the duplicate tests too, then this can go. I'll do the necessary rebases, no worries.

Hmm, why were you expecting it to not pass, and why would the other PR fix that? Not going to merge since you expect it isn't ready.

dcs/coalition.py
332fg-raven: @DanAlbert - I don't think so or I am not seeing something. As I understand d["country"] is a dict of countries with units on the map. The block which includes L259 goes through all countries with units on the map and creates the respective country objects (and their units) and adds them to the coalition. The block I added goes over the mission's coalitions dict, which is essentially a list of country_ids belonging to coalition - and does not have any unit information. So add_country would be called twice for countries already processed, effectively deleting all their units since countries.get_by_id() creates a new object. Right? I'll update the mission file to have 1 unit in one of the countries and test to validate that loaded and saved missions both contain the same amount of units. Let me know if you see it differently.

332fg-raven:
@DanAlbert - mypy tests/test_mission.py output on maser branch run in my local env says the following:
```
tests/test_mission.py:1007: error: Name "test_big_formation_action_leader" already defined on line 891  [no-redef]
tests/test_mission.py:1036: error: Name "test_big_formation_action_left" already defined on line 920  [no-redef]
tests/test_mission.py:1065: error: Name "test_big_formation_action_back" already defined on line 949  [no-redef]
tests/test_mission.py:1094: error: Name "test_big_formation_action_right" already defined on line 978  [no-redef]
```

dcs/coalition.py
DanAlbert: Oh, I see what's going on. This is a little odd but it's good enough. I don't expect this is ever going to be a performance issue, but I thought the optimization was trivial rather than a minor refactor.

332fg-raven:
@DanAlbert  - I see, fyi, mypy that I'm using:
```
$ mypy --version
mypy 1.7.1 (compiled: yes)
```

dcs/coalition.py
332fg-raven: @DanAlbert - agreed, initial patch had it in a single loop, but I think your suggestion to keep it outside that loop makes more readable since it's focusing on just countries without units on the map.

332fg-raven:
@DanAlbert - if mypy issues are not an issue, this is good to be merged. I've updated the test which makes sure that there's 1 unit in the "UK" after loading/saving and renamed .miz file since it's not empty (of units) anymore.

332fg-raven:
@DanAlbert - do I have to update anything on this or? From my POV it's ready to be merged.

DanAlbert:
I'll get to it when I have time. I haven't for the past few days.

332fg-raven:
> I'll get to it when I have time. I haven't for the past few days.

No issues, was just wondering if I missed anything and you were waiting for it.

PR #356, opened 2024-01-15
Support for CarpetBombing task

332fg-raven:
This change allows WWIIBigFormation to use Carpet Bombing task.
Removed duplicate tests for big formation which crept in with PR #351


dcs/task.py
```diff
@@ -162,8 +162,8 @@ def stop_after_duration(self, duration: int):
         self.params["stopCondition"]["duration"] = duration
 
 
-class WeaponType(Enum):
-    Auto = 1073741822
+class WeaponType(IntEnum):
+    Auto = 9663676414
```
DanAlbert, line 166: This is sort of alarming, but I created a test miz ([weapon type.zip](https://github.com/pydcs/dcs/files/13942649/weapon.type.zip)) and it agrees. I wonder what DCS has been doing with the wrong value...

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+
+    class AttackType(Enum):
+        Carpet = "Carpet",
+
+        def __eq__(self, other: Any) -> bool:
+            if isinstance(other, str):
+                return self.value == other
+            return self == other
+
+    def __init__(self, altitude: int = 2000, altitude_enabled: bool = False,
```
DanAlbert, line 1096: `altitude` is a misleading name. It's not the altitude for the attack, it's the minimum allowed altitude.

Replace `altitude` and `altitude_enabled` with a single `minimum_attack_altitude: Optional[int] = None`

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+                 attack_qty: int = 1, attack_qty_limit: bool = False,
```
DanAlbert, line 1097: This doesn't appear to be an option in the ME. What is this?

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+                 attack_type: AttackType = AttackType.Carpet, carpet_length: int = 500,
```
DanAlbert, line 1098: `attack_type` also doesn't appear to be an option in the ME. What is this?

DanAlbert, line 1098: s/`carpet_length`/`pattern_length_ft`/

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+                 expend: Expend = Expend.Auto, group_attack: bool = False,
```
DanAlbert, line 1099: Group attack doesn't appear to be an option in the ME.

(I'm looking at Perform Task -> Carpet bombing, and have looked at B-52, B-1, and B-17, all of which have the same options)

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+                 weapon_type: WeaponType = WeaponType.Auto,
+                 x: float = 0, y: float = 0) -> None:
```
DanAlbert, line 1101: Also don't appear in the ME.

dcs/task.py
```diff
@@ -1077,6 +1082,39 @@ def __eq__(self, other) -> bool:
...
+class CarpetBombing(Task):
+    Id = "CarpetBombing"
+
+    class AttackType(Enum):
+        Carpet = "Carpet",
+
+        def __eq__(self, other: Any) -> bool:
+            if isinstance(other, str):
+                return self.value == other
+            return self == other
```
DanAlbert, line 1094: Does this not just infinitely recurse?

```python
>>> class Foo(Enum):
...     A = "a"
...     def __eq__(self, o):
...             if isinstance(o, str):
...                     return self.value == o
...             return self == o
...
>>> Foo.A == 0
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 6, in __eq__
  File "<stdin>", line 6, in __eq__
  File "<stdin>", line 6, in __eq__
  [Previous line repeated 496 more times]
RecursionError: maximum recursion depth exceeded
```

DanAlbert (changes requested):
You've got 3 unrelated PRs in this PR:

1. Add CarpetBombing
2. Fix an incorrect Auto value
3. Delete a duplicate test

I could have merged 2/3 of those if they'd been distinct PRs. It makes things go a lot more smoothly when PRs are split up.

dcs/task.py
```diff
@@ -323,6 +323,11 @@ class Expend(Enum):
     Half = "Half"
     All = "All"
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, str):
```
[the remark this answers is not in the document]
DanAlbert: Disregard anyway, I put this on the wrong line :)

dcs/task.py
DanAlbert: If you can't explain what it does, don't put it in the API. The implementation can fill this in, but I've got no idea what this does (and it sounds like you don't either), so it's not a useful parameter.

I'd recommend creating a mission using this task, look at the lua, and hard code whatever it uses. Make a few of them with different settings for the weapon type, probably, and make sure they're all the same.

If the value is not constant and you can't figure out a pattern, I've got no idea what to do here, but punting that question to the caller isn't going to work.

dcs/task.py
DanAlbert: Yes, sorry, I meant to say meters, but was reading the ME when I typed it and mine is also configured for US units.

dcs/task.py
DanAlbert: If you can't explain their behavior we shouldn't expose them. We might be doing more harm than good by exposing them now. We don't want a DCS update to expose latent bugs in callers that were setting a previously ignored property.

dcs/task.py
DanAlbert: I'd be able to merge that now rather than whenever the carpet bombing task is ready.

dcs/task.py
DanAlbert: > Improving python skills at every step

In case it helps, `a == b` calls `a.__eq__(b)`. If that returns `True` or `False`, that's the end of it. If it returns `NotImplemented` (not to be confused with raising `NotImplementedError`, which is something else), it retries with `b.__eq__(a)` (that last part I actually only learned the other day while looking through the docs for reviewing your other PR, so you'll probably find a bunch of existing implementations here that don't do that).

What you've got here is `a.__eq__(b)` which in turn calls `a.__eq__(b)` unless `b` is an instance of `str` :)

dcs/task.py
332fg-raven: @DanAlbert - I understand and thinking about it now (after understanding more about DCS/miz/pydcs) I think ME should be preferred since ME user is only capable of changing the ME things and they should be named following ME naming convention too, since pydcs caller is a user emulating ME and object parameters in pydcs are the ME parameters. I'll refactor this.

dcs/task.py
332fg-raven: @DanAlbert - sumitted as a new PR.

dcs/task.py
332fg-raven: @DanAlbert - I simplified the implementation to care only about str and Expend instances, others are not equal.

332fg-raven:
@DanAlbert - merge PR #357 first, I'll rebase this and then the tests here will pass.

332fg-raven:
@DanAlbert - rebased, ready to be merged.

dcs/task.py
```diff
@@ -323,6 +323,13 @@ class Expend(Enum):
...
+            return self.value == other
+        if isinstance(other, Expend):
```
DanAlbert, line 329: Close enough so meh, non-blocking, but for the future just `return super().__eq__(other)` after doing any custom checks.

dcs/task.py
```diff
@@ -1077,6 +1084,38 @@ def __eq__(self, other) -> bool:
...
+
+    def __init__(self, alt_above: int = DEFAULT_ALT,
+                 pattern_length: int = 500,
+                 release_qty: Expend = Expend.Auto,
+                 weapon_type: WeaponType = WeaponType.Auto) -> None:
+        super().__init__(self.Id)
+        # hardcoded parameters are present in .miz file, but not visible in Mission Editor
+        self.params = {
+            "altitude": alt_above,
+            "altitudeEnabled": True if alt_above != CarpetBombing.DEFAULT_ALT else False,
```
DanAlbert, line 1106: I don't think this is actually right. I suspect this field is supposed to be "did the caller make a choice, or did they use the default?". This can't distinguish between a chosen altitude of 2000 and a default behavior. For that you need `alt_above: int | None = None`.

I've also got no idea what the impact of getting this wrong is. Might be just UI behavior in the ME? I won't block on this for now, but for future reference, that's how this thing ought to be written.

DanAlbert: Also, fyi, this is just a more verbose way to write `"altitudeEnabled": alt_above != CarpetBombing.DEFAULT_ALT`

DanAlbert (approved):

PR #360, opened 2024-01-19
Fix for tmpdir when loading multiple mission files

332fg-raven:
This change makes sure that each mission file, when loaded, it extracts resources to unique temp dir rather than a shared one, which would lead to resource overwriting when working with multiple loaded missions.
I worked with two missions, and each had "briefing.png" attached. When I saved them I noticed that one was overwriting the other.

I guess accessing map_resource.files["DEFAULT"].values() directly might not be desired and if you want, I can remove it. I think if m.tmpdir is good, then a validation that m1.tmpdir != m2.tmpdir should be enough.

NOTE that get_file_path works only when .miz is loaded. It does not work properly (even before this patch) for missions generated by pydcs, where resources are added, e.g:

m = dcs.Mission(terrain=dcs.terrain.Caucasus())
res_key = m.add_picture_blue("tests/images/blue")
calling m.map_resource.get_file_path(res_key) would returns "'cs/dcs/tests/images/blue.png'". But that's a separate problem to solve.

With this patch the get_file_path() throws an exception, which is IMHO better than silently working wrong.

tmpdir is optional since I didn't see the need of creating a new tmp folder every time a new Mission() object is created.

Let me know your thoughts.


DanAlbert:
It's definitely an incremental improvement, so I'm inclined to just merge as-is. That file path API implementation is weird though (as you say, a pre-existing condition). I don't know how it's intended to be used though so I'm not sure if it's right or wrong.

PR #362, opened 2024-01-28
Support for a_out_text_delay_u action

332fg-raven:


dcs/action.py
```diff
@@ -923,6 +923,36 @@ def dict(self):
         return d
 
 
+class MessageToUnit(TextAction):
+    predicate = "a_out_text_delay_u"
+
+    def __init__(self, unit_id: int = 0, text: String = String(),
```
DanAlbert, line 929: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

DanAlbert (changes requested):

dcs/action.py
332fg-raven: @DanAlbert - not sure I understand the issue here. I copied the code from other MessageToX actions. text arg is not being modified (appended, deleted, etc). It's true that default value is empty string, initialized at parse time, but that should be allright, no?

332fg-raven: @rp- do I need to fix anything here or not? The text variable is not modified anywhere, so I guess the mutable-default-args is not that much an issue or am I not seeing something?

rp-: Yes it will not be an issue here, but for cleaness you should probably rewrite it to:
```python
text: String = None

tmp_text = String() if text is None else text
super().)__init__(self.predicate, tmp_text)
....
self.params.append(tmp_text)
```

332fg-raven: @rp- I see, should we follow such approach even when class such as MessageToUnit does not offer any interfaces to change the text later? In my view, optional parameters are optional, text in this case is not. I've seen interfaces where everything is optional and that gets hard to maintain. And the fact that the String object is passed via mission.translation.get_string(d["text"]) and then having it defined as optional just confuses the reader. I'm trying to understand and improve - that's why I'm asking.
If you say tmp_text is the way, I'll do it, no issues, just clarifying why I think it's a bad case to do it.

rp-: As I/we did it "wrong" in the whole action.py it is fine to keep it like it is.
And if we ever intend to make this somehow mutable we have to cleanup all occurrences anyway.

PR #363, opened 2024-01-29
Make it possible to deserialize missions containing RecoveryTanker tasks

magwo:


magwo:
Not sure if we need None handling of the lastWaypoint thing.

SnappyComebacks:
Thanks for catching the missing entry.

dcs/task.py
```diff
@@ -739,7 +739,7 @@ def __init__(self):
 class RecoveryTanker(Task):
     Id = "RecoveryTanker"
 
-    def __init__(self, groupId: int, speed: float, altitude: float, lastWaypoint: Optional[int]):
+    def __init__(self, groupId=None, speed=600, altitude=4000, lastWaypoint: Optional[int] = None):
```
DanAlbert, line 742: Don't delete the type annotations. You've changed this signature from `float` to `int` here.

DanAlbert:
> Also missing default parameters which I think is preventing deserialization.

Isn't that because it's missing a `create_from_dict` override? That seems the better fix, unless there's a good reason why one of these tasks should be creatable with no group ID.

(making the optional last waypoint argument optional makes sense, of course)

magwo:
Good points, will take a look.

PR #367, opened 2024-02-08
Unit Group password support

332fg-raven:


dcs/unitgroup.py
```diff
@@ -223,8 +228,27 @@ def dict(self):
...
+        SALT_SIZE = 11
+        DIGEST_SIZE = 32
+
+        key = ''.join(random.sample(string.digits + string.ascii_letters, SALT_SIZE))
+        p_hash = hashlib.blake2b(key=key.encode(), digest_size=DIGEST_SIZE)
+        p_hash.update(password.encode())
+        b64hash = base64.encodebytes(p_hash.digest()).decode()
+
+        # remove '=\n' at the end.
+        self.password = key + ":" + b64hash[:len(b64hash) - 2]
```
DanAlbert, line 247: You don't need `len(b64hash)` here. Just `-2` does the same thing.

PR #368, opened 2024-02-13
Fix for setting password for Unit Groups

332fg-raven:
* Password is not base64 encoded, but base64URL encoded. Missed that in the initial read.

### DIFF
--- a/dcs/password.py
+++ b/dcs/password.py
@@ -1,0 +1,38 @@
+"""
+The password module generates DCS compliant passwords from input string
+"""
+
+import hashlib
+import base64
+import string
+import random
+
+
+def password_hash(password: str) -> str:
+    # see https://www.reddit.com/r/hoggit/comments/uf2sh0/psa_creating_the_new_slot_passwords_outside_of_dcs/
+
+    # encode from https://github.com/simonwhitaker/base64url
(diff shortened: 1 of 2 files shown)


332fg-raven:
@rp- or @DanAlbert - can you please merge this. If there's anything wrong, let me know so I can address it. Thank you!

dcs/__init__.py
```diff
@@ -20,5 +20,6 @@
 from . import forcedoptions
 from . import installation
 from . import nav_target_point
+from . import password
```
rp-, line 23: not sure if we need that in the main package domain?
I think it would be enough to just import it in `unitgroup.py`

332fg-raven: @rp- the reason I added password as a separate module is due to the fact that I'm working on another patch which adds support for ground control passwords. And it felt wrong to import it from unitgroup. If you think it's better to keep it in unitgroup, I can definitely import it that way.

rp-: But I guess the password module will just used internally?
So there will be no reason to "export" the module in the main module.

332fg-raven: @rp- noted. Updated.

PR #373, opened 2024-02-24
Support miz files saved by old versions of pydcs where not all countries in a coalition were saved

zhexu14:
This PR changes mission loading behavior to support .miz files saved by legacy versions of pydcs  (prior to #355 )  where countries without units were not saved. The DCS ME supports loading these files, so pydcs should probably support these files to the best extent it can. AFAICT the change introduced in this PR means that the legacy behavior is maintained for legacy .miz files.

### DIFF
--- a/dcs/mission.py
+++ b/dcs/mission.py
@@ -358,8 +358,13 @@ class Mission:
         for col_name in ["blue", "red", "neutrals"]:
             if col_name in imp_mission["coalition"]:
                 self.coalition[col_name] = Coalition(col_name, imp_mission["coalition"][col_name]["bullseye"])
+                # Support .miz files saved by earlier versions of DCS or pydcs where countries in coalition 
+                # were not saved.
+                countries_in_coalition = {}
+                if col_name in imp_mission["coalitions"]:
+                    countries_in_coalition = imp_mission["coalitions"][col_name]
                 status += self.coalition[col_name].load_from_dict(self, imp_mission["coalition"][col_name],
-                                                                  imp_mission["coalitions"][col_name])
+                                                                  countries_in_coalition)
 
         # triggers
         self.bypassed_triggers = None


Raffson:
It seems that one of misunderstood the root cause of the issue here.

If I understood you correctly, you're saying that older miz files don't save countries that have no units under the coalition section.

I concluded something different though, i.e. older missions used to serialize the neutral coalition using "neutral" as key instead of "neutrals". You can open one of the failing layouts and see how the coalitions are serialized, and that's where a difference can be spotted compared to new missions. So now I'm really wondering how you concluded that earlier missions didn't serialize countries without units 😅

Nevertheless, I still think this solution will work for Liberation, as long as none of the layouts are defined using a neutral coalition, but I don't think that's the case. Personally I opted for an approach where you check for the presence of "neutrals", and if it isn't, try "neutral" instead. This way you still load in the neutral coalition from the mission.

zhexu14:
> It seems that one of us misunderstood the root cause of the issue here.
> 
> If I understood you correctly, you're saying that older miz files don't save countries that have no units under the coalition section.
> 
> I concluded something different though, i.e. older missions used to serialize the neutral coalition using "neutral" as key instead of "neutrals". You can open (in a text editor) one of the failing layouts and see how the coalitions are serialized, and that's where a difference can be spotted compared to new missions. So now I'm really wondering how you concluded that earlier missions didn't serialize countries without units 😅
> 
> Nevertheless, I still think this solution will work for Liberation, as long as none of the layouts are defined using a neutral coalition, but I don't think that's the case. Personally I opted for an approach where you check for the presence of "neutrals", and if it isn't, try "neutral" instead. This way you still load in the neutral coalition from the mission.

Good call outs. My PR message was probably a bit unclear, the reference to countries with no units came from the wording in #355 which introduced the new behavior in pydcs. I'm in agreement that the `neutral` vs `neutral` is the root cause.

On solving the issue, my 2 cents is that pydcs should behave in the same way the DCS Mission Editor behaves. If the Mission Editor can load a .miz file, so should pydcs, and the two should interpret the legacy .miz files the same way. If this causes problems with Liberation, then the fix should be to update the .miz files in Liberation.

However, you're right in that `neutral` is indeed in the .miz files and I've marked this PR as "Draft" until I can go into the DCS Mission Editor and see what it does with the legacy files.

zhexu14:
@Raffson , in case you were interested, I dug into the DCS Mission Editor and found that at the moment at least, the DCS ME doesn't seem to use the neutral country list. Instead, it seems to populate the neutral list by whatever countries aren't in red or blue, so it's safe to just save nothing into the neutrals county list and let DCS figure it out :)

Raffson:
@zhexu14 I believe that's one of the built-in contingencies of the ME, i.e. of the scripts parses the miz file and corrects where necessary. So I suspect the parsing of "neutrals" _could_ be dropped, but it _could_ also cause issues for dedicated servers because those have no "safety nets", i.e. the dedicated server runs the mission exactly how it receives it without making any changes like the ME would apply. Does that make sense? 😅

zhexu14:
> @zhexu14 I believe that's one of the built-in contingencies of the ME, i.e. of the scripts parses the miz file and corrects where necessary. So I suspect the parsing of "neutrals" _could_ be dropped, but it _could_ also cause issues for dedicated servers because those have no "safety nets", i.e. the dedicated server runs the mission exactly how it receives it without making any changes like the ME would apply. Does that make sense? 😅

Interesting -- do you know if the dedicated server runs missions differently from running missions normally? I've never tried multiplayer / using the dedicated server but will have a crack at running a server when I get a chance. 

FWIW I did test running the mission both from the Mission Editor and from the Missions screen (i.e. not through the ME) and the handling of neutral units is identical. What you say about the Mission Editor having "safety nets" makes sense and tracks with my experience running missions through the Missions screen, which can sometimes behave differently from starting the mission via the ME.

Raffson:
Afaik the safeties will apply when you start the mission through the client in general. But as soon as you run the mission through a dedicated instance, any serialization issues _could_ result into unexpected behavior.

Meanwhile it's perfectly possible that in this particular case the dedicated server will actually determine the coalitions at runtime rather than relying on the mission file, because it can deduce it from the units that are defined since they're defined per coalition.

All of that said, I don't think any of this would actually apply to Liberation or any other pydcs user _unless_ they're trying to deserialize a pretty old mission, in which case they can fix it by opening the ME and re-saving it without any changes. Also, Liberation will inject all neutral countries during mission generation (unless that changed) so I think this solution will work just fine unless you want to define layouts with neutral units, but that's not even supported by Liberation afaik so no worries there. The last thing that comes to mind is probably the campaign's miz file itself, but iirc it doesn't rely on neutral units either 😅

PR #377, opened 2024-04-20
renamed flag_black to proper name flag_back

pschiel:
resolves error when mission contains a wait-for-user action using "flag_back"


rp-:
Thanks

PR #378, opened 2024-04-21
Support for find_unit and find_unit_by_id

332fg-raven:


dcs/mission.py
```diff
@@ -1844,6 +1844,56 @@ def find_group_by_id(self, group_id: int) -> Optional[Group]:
...
+            coalition: coalition to search in
+
+        Returns:
+            Unit: the unit found, otherwise None
+        """
+
+        coalitions = [coalition] if coalition is not None else self.coalition.values()
+        for k in coalitions:
+            for c in k.countries.values():
+                groups: List[Sequence[Group]] = [c.ship_group, c.vehicle_group, c.plane_group,
```
rp-, line 1862: Wouldn't it be simpler to just concat the group lists? instead of creating a list of groups?
e.g.:
```
groups: List[Group] = c.ship_group + c.vehicle_group....
```

Then it would be 1 for loop less

dcs/mission.py
```diff
@@ -1844,6 +1844,56 @@ def find_group_by_id(self, group_id: int) -> Optional[Group]:
...
```
rp-, line 1887: same as above

dcs/mission.py
332fg-raven: @rp-  - no issues, will do it.

332fg-raven:
@rp- If I use List[Group] on groups, mypy shows errors:
```
dcs/mission.py:1862: error: Incompatible types in assignment (expression has type "list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]", variable has type "list[Group[Any, Any]]")  [assignment]
dcs/mission.py:1862: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
dcs/mission.py:1862: note: Consider using "Sequence" instead, which is covariant
dcs/mission.py:1862: error: Unsupported operand types for + ("list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]" and "list[HelicopterGroup]")  [operator]
```

If I use Sequence[Group] as suggested, I get the following:
`dcs/mission.py:1862: error: Unsupported operand types for + ("list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]" and "list[HelicopterGroup]")  [operator]`

I could use `list[HelicopterGroup | StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]` or create a new alias for that.
Using none, does not give any warnings. Let me know if you want to keep it like that or should I revert back.

rp-:
> @rp- If I use List[Group] on groups, mypy shows errors:
> 
> ```
> dcs/mission.py:1862: error: Incompatible types in assignment (expression has type "list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]", variable has type "list[Group[Any, Any]]")  [assignment]
> dcs/mission.py:1862: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
> dcs/mission.py:1862: note: Consider using "Sequence" instead, which is covariant
> dcs/mission.py:1862: error: Unsupported operand types for + ("list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]" and "list[HelicopterGroup]")  [operator]
> ```
> 
> If I use Sequence[Group] as suggested, I get the following: `dcs/mission.py:1862: error: Unsupported operand types for + ("list[StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]" and "list[HelicopterGroup]") [operator]`
> 
> I could use `list[HelicopterGroup | StaticGroup | ShipGroup | VehicleGroup | PlaneGroup]` or create a new alias for that. Using none, does not give any warnings. Let me know if you want to keep it like that or should I revert back.

I'm surpised mypy can't figure this out, until we have something wrong with the inheritance, but they all share the `Group` as base type.
But I'm fine with specifying nothing, as it is pretty clear to the reader I guess.

PR #380, opened 2024-04-27
Support for linked trigger zones

332fg-raven:


332fg-raven:
@rp- any particular issue with this merge?

PR #382, opened 2024-05-25
DCS Export version 2.9.5.55300

zhexu14:


dcs/statics.py
```diff
@@ -462,6 +462,618 @@ class Beer_Bomb(unittype.StaticType):
...
+        id = "Container_20ft"
+        name = "M92 Container 20ft"
+        shape_name = "M92_Container_20ft"
+        rate = 1
+
+    class Container_40ft(unittype.StaticType):
+        id = "Container_40ft"
+        name = "M92 Container 40ft"
+        shape_name = "M92_Container_40ft"
+        rate = 1
```
Raffson, line 679: These two are conflicting with existing "containers". Notice how the "id" is different is casing compared to the ones starting at line 1167.
At any rate, one way to fix this is is to avoid capitalizing the first letter of the class, but that'll change quite a few other vehicles as well which _can_ cause a headache when updating pydcs in Liberation 😬

zhexu14: Yeah, I saw your commits in the Retribution fork of pydcs. I went a different way of suffixing duplicates with an underscore, which means there isn't any change to existing unit classes. Probably doesn't make a huge difference for Liberation / Retribution but at the end of the day pydcs is a library and keeping class names unchanged (as much as DCS data allows) would be good.

Raffson: Agreed, that's definitely the better option in hindsight 😅

PR #387, opened 2024-08-01
Fix spawns on carriers and FARPs in DCS 2.9.6

zhexu14:
With DCS2.9.6, a warehouse is needed for aircraft to spawn on carriers and FARPs. This change in DCS means the following trivial example missions would fail:

```
from dcs.mission import Mission
from dcs.ships import Stennis, HarborTug
from dcs.planes import FA_18C_hornet
from dcs.helicopters import Mi_8MT
from dcs import countries  
from dcs import mapping
from dcs import unit


mission = Mission()
blue = mission.country("USA")
sg = mission.ship_group(blue, "Carrier", Stennis, mapping.Point(-200000, 450000, mission.terrain))
ag = mission.flight_group_from_unit(blue, "Player", FA_18C_hornet, sg)
ag.units[0].skill = unit.Skill.Player
mission.save("carrier_spawn.miz")

mission = Mission()
blue = mission.country("USA")
farp = mission.farp(blue, "FARP", mapping.Point(-284879, 636819, mission.terrain))
ag = mission.flight_group_from_unit(blue, "Player", Mi_8MT, farp)
ag.units[0].skill = unit.Skill.Player
mission.save("farp_spawn.miz")

```

This PR introduces a function to update warehouses for units in a mission and calls this function before saving a mission to a .miz file. This PR also refactors how warehouses are represented and adds new mandatory Warehouse fields.

### DIFF
--- a/dcs/mission.py
+++ b/dcs/mission.py
@@ -1006,7 +1006,7 @@ class Mission:
             ptask.auto = True
             mp.tasks.append(ptask)
         return mp
-        
+
     def update_warehouses(self):
         """Some units need to have warehouse entries. This function updates warehouse entries based on units defined
         in the mission.
--- a/dcs/mission.py
+++ b/dcs/mission.py
@@ -673,6 +673,7 @@ class Mission:
         sg.add_point(sp)
 
         country.add_static_group(sg)
+        self.warehouses.warehouses[str(sg.id)] = terrain_.terrain.Warehouse().dict()
         return sg
 
     def vehicle(self, name: str, _type: Type[unittype.VehicleType]) -> Vehicle:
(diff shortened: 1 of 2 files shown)


magwo:
@zhexu14 What was the symptoms when carrier spawns wasn't working?

Are you sure this change fixes it? We have a tool the loads a miz, manipulates the weather randomly, and then saves a new miz file.

After using out using with the latest pydcs version (containing this commit), it doesn't appear to work. When trying to spawn on a carrier there is just a slight pause for a second and then nothing happens.

Said tool can be found [here](https://github.com/magwo/dcs-mission-buzzer).

magwo:
When you wrote:
> the following trivial example missions would fail

What did you mean? Fail in what manner?

zhexu14:
> When you wrote:
> 
> > the following trivial example missions would fail
> 
> What did you mean? Fail in what manner?

It fails in the sense that you would expect the missions generated by the code snippet to spawn you into DCS into the aircraft (F/A-18 in the carrier spawn, Huey in the FARP spawn) when loading the mission via the "Missions" menu in DCS i.e. without going via the Mission Editor. Without this PR what actually happens in DCS 2.9.6 is that you spawn in 3rd person view and the aircraft never appears. 

Without this PR, loading the mission via the Mission Editor and saving would result in the expected the behavior.

zhexu14:
> @zhexu14 What was the symptoms when carrier spawns wasn't working?
> 
> Are you sure this change fixes it? We have a tool the loads a miz, manipulates the weather randomly, and then saves a new miz file.
> 
> After using our tool (using the latest pydcs version (containing this commit)), it doesn't appear to work. When trying to spawn on a carrier there is just a slight pause for a second and then nothing happens.
> 
> Said tool can be found [here](https://github.com/magwo/dcs-mission-buzzer).

What you're describing sounds like the issue this PR is trying to fix. 

Did dcs-mission-buzzer work on DCS2.9.6 without this commit? Just trying to understand if this PR made things go backwards or if it was just not a complete fix.

Also, how did you generate the mission that was loaded into dcs-mission-buzzer? Was it via the DCS Mission Editor or "from scratch" using pydcs?

Issue #427, opened 2026-05-21
`Mission.update_warehouses()` overwrites existing ship/FARP warehouse entries on every save


MacrayBlackhand:
### Summary

`update_warehouses()`, called automatically by `Mission.save()`, unconditionally
replaces the warehouse entry for every parking-capable ship and every FARP with
a fresh default `Warehouse`. Any warehouse configuration applied to a carrier or
FARP — operating levels, fuel init, `dynamicCargo`, inventory, suppliers — is
silently discarded at save time.

### Affected code

`dcs/mission.py`, method `Mission.update_warehouses()` (around line 1018).

The method iterates parking-capable ships and `BaseFARP` statics and does, for
each one:

```python
self.warehouses.warehouses[int(unit_.id)] = terrain_.terrain.Warehouse().dict()
self.warehouses.warehouses[int(unit_.id)]['coalition'] = coalition_name_in_warehouse
```

This assignment is unconditional — it runs whether or not an entry for that
unit id already exists.

### Why this is a bug

The method's own docstring describes its purpose as:

> "Some units need to have warehouse entries. This function updates warehouse
> entries based on units defined in the mission."

The intent expressed there is to *ensure entries exist* (gap-fill). The actual
behavior is to *overwrite* — every existing entry is destroyed and replaced
with a default on every save. Because `save()` always calls
`update_warehouses()`, there is no way for a caller to persist any custom
warehouse state on a ship or FARP: it is reset immediately before the file is
written.

### Reproduction

```python
import dcs

m = dcs.Mission()
# ... add a parking-capable carrier to the mission, note its unit id ...

# configure that carrier's warehouse
wh = m.warehouses.warehouses          # after the unit exists / first save
# set e.g. unlimitedFuel = False, OperatingLevel_Air = 3, dynamicCargo = True

m.save("test.miz")
# update_warehouses() runs inside save() and resets the carrier's entry
# to a fresh default before the file is written.

m2 = dcs.Mission()
m2.load_file("test.miz")
# the carrier's warehouse is back to defaults — custom values are gone
```

### Expected behavior

`update_warehouses()` should be gap-fill only: create a default entry for a
ship/FARP that does not yet have one, and leave any existing entry untouched.

### Suggested fix

Guard each write so it only creates a missing entry:

```python
uid = int(unit_.id)
if uid not in self.warehouses.warehouses:
    self.warehouses.warehouses[uid] = terrain_.terrain.Warehouse().dict()
[...]

PR #436, opened 2026-06-12
Read warehouses in Warehouses.load_dict (symmetric with serialization)

### LINKED ISSUE #428: `Warehouses.load_dict()` does not read unit (ship/FARP) warehouse entries; save/load is asymmetric
### Summary

`Warehouses.load_dict()` reads only the `airports` section of the warehouses
data. The `warehouses` section — which holds ship and FARP warehouse entries
keyed by unit id — is never read. `Warehouses.__str__()` *does* write that
section, so unit warehouse state is serialized into the `.miz` correctly but is
silently dropped on load.

### Affected code

`dcs/terrain/terrain.py`, class `Warehouses` (around line 609).

The load side:

```python
def load_dict(self, data):
    for x in data.get("airports", {}):
        self.terrain.airport_by_id(x).load_from_dict(data["airports"][x])
```

The save side (`__str__`):

```python
def __str__(self):
    airports = self.terrain.airports
    d = {
        "warehouses": self.warehouses,
        "airports": {airports[x].id: airports[x].dict() for x in airports}
    }
    return lua.dumps(d, "warehouses", 1)
```

`__str__()` writes both `warehouses` and `airports`. `load_dict()` reads only
`airports`. The `warehouses` key is ignored on load.

### Why this is a bug

The save and load paths are asymmetric. Unit warehouse data (ships with
parking, FARPs) is written into the `warehouses` block of the saved `.miz` but
is never restored into `self.warehouses` when the mission is loaded. After
`Mission.load_file()`, `Warehouses.warehouses` is empty regardless of what the
file contained, so any ship/FARP warehouse state present in the file is lost.

### Reproduction

```python
import dcs

m = dcs.Mission()
# ... add a parking-capable ship, give its warehouse entry some non-default
#     values directly in m.warehouses.warehouses[unit_id] ...
m.save("test.miz")
# inspect test.miz: the 'warehouses' Lua file contains a 'warehouses' table
# with the unit entry. The data is present on disk.

m2 = dcs.Mission()
m2.load_file("test.miz")
print(m2.warehouses.warehouses)   # {} — the unit entries were not read back
```

### Expected behavior

`load_dict()` should restore the `warehouses` section symmetrically with what
`__str__()` writes, so unit warehouse entries present in the file are available
on `self.warehouses.warehouses` after load.

### Suggested fix

Add the missing read of the `warehouses` section:

```python
def load_dict(self, data):
    for x in data.get("airports", {}):
        self.terrain.airport_by_id(x).load_from_dict(data["airports"][x])
    self.warehouses = data.get("warehouses", {})
```

### Related

Compounds with Issue 1 (`Mission.update_warehouses()` overwrites existing
ship/FARP warehouse entries on save). Both fixes are required for unit
warehouse state to survive a full save/load round-trip.

MacrayBlackhand:
Warehouses.load_dict reads only data["airports"] and silently drops
data["warehouses"], so unit/FARP/ship warehouse state is lost on every
save->load round-trip even though the serializer writes it. This is the read-side
mirror of the write-side gap in #427 -- together, unit warehouse state survives
neither save nor reload.

Read the warehouses sub-dict symmetrically with the airports loop.

Adds a round-trip test that loads a payload with a warehouse entry and asserts it
survives.

Closes #428.

### DIFF
--- a/dcs/terrain/terrain.py
+++ b/dcs/terrain/terrain.py
@@ -614,6 +614,8 @@ class Warehouses:
     def load_dict(self, data):
         for x in data.get("airports", {}):
             self.terrain.airport_by_id(x).load_from_dict(data["airports"][x])
+        for uid, wh_data in data.get("warehouses", {}).items():
+            self.warehouses[int(uid)] = wh_data
 
     def __str__(self):
         airports = self.terrain.airports
--- a/tests/test_terrain.py
+++ b/tests/test_terrain.py
@@ -94,6 +94,24 @@ class NevadaTest(unittest.TestCase):
         slots = m.terrain.airports["Nellis"].free_parking_slots(dcs.planes.KC_135)
 
 
+class WarehousesTest(unittest.TestCase):
+
+    def test_load_dict_reads_warehouses(self):
+        # __str__ serializes both "airports" and "warehouses", but load_dict
+        # historically read only "airports", silently dropping unit/FARP/ship
+        # warehouse state on every save->load round-trip.  load_dict must read
+        # "warehouses" symmetrically with the serializer.
+        m = dcs.mission.Mission(terrain=dcs.terrain.Caucasus())
+        m.warehouses.load_dict({
+            "airports": {},
+            "warehouses": {"4242": {"coalition": "BLUE", "size": 100}},
+        })
+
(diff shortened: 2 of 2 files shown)


si0phia (approved):
I see no conflicts from these line changes

Issue #317, opened 2023-05-19
replace our lua parser with lupa?


DanAlbert:
https://github.com/pydcs/dcs/pull/316 was needed to work around a parsing error in a new livery in today's update.

I also fixed about a dozen other parser errors the other week. Should we just take the dependency on lupa, continuing to squash bugs in our own parser as they come up? Worth mentioning that even with lupa we won't be bulletproof, since parsing can easily be broken by DCS deciding to change the format or even just doing things like `name = identity("foo")` (though if we had lupa we would at least be able to evaluate those, but we'd need to do so proactively).

@rp-, it's your call. My opinion doesn't count for much because adding lupa would be a hardly noticeable change for liberation, but might be for smaller projects.

rp-:
I think my reasoning against lupa back than, was introducing that dependency.
And at first it was just about parsing the relative simple .miz "lua" format.

I guess since we introduced `pyproj`, there is not much reasoning against another bigger dependency like this.

DanAlbert:
Yeah, when it's just the Miz we can get away with a lot less since we only have to cope with whatever the ME produces. For liveries and a handful of other things we may want to parse where they're hand written we'd actually need a complete parser :)

I'm not sure if I'm going to look into this any time soon, but that's the plan I'll go with next time something breaks, or if I decide I want access to those few liveries that we can't currently parse.

DanAlbert:
It's exactly that. We don't assign TOTs for nav points, and pydcs apparently defaults all waypoint speeds to 600kph, which is really quite slow.

Maintenance-Partnership-Systems:
Just to add support for this suggestion - I found this issue because I was diagnosing a crash in Liberation when loading custom liveries.  The cause was the use of [[ ... ]] quoted strings for the file paths in the description.lua, and the home-grown parser doesn't handle that.  A general-purpose parser might be more likely able to handle unusual cases like this.

Barabas5532:
I'm looking at the F-100D UnitPayloads Lua file and it uses a function to generate the loadout data. It also uses alternate syntax (`unitType = "F-100D",`) for defining tables that break a regex used for finding this value. Using a real interpreter could help with this module.

PR #440, opened 2026-06-16
load_payloads: tolerate globbed-but-uncached payload files

MacrayBlackhand:
## Problem

FlyingType.load_payloads() aborts the entire mission load with a KeyError
when a payload directory contains a *.lua file that scan_payload_dir()
globbed but did not cache.

The two halves of the payload machinery use mismatched guards:

- scan_payload_dir() globs every *.lua in each payload dir, but only inserts
  a cache entry when the file content matches the regex
  `\["unitType"]\s*=\s*"([^"]*)` (dcs/unittype.py, ~L108-122). A file whose
  unitType is written in a form the regex doesn't match is globbed but never
  cached.
- load_payloads() then re-globs the same *.lua files and reads the cache with
  a bare subscript (dcs/unittype.py:140):

```python
  if FlyingType._payload_cache[payload_path] == cls.id and payload_path.exists():
```

  For any globbed-but-uncached path, _payload_cache[path] raises KeyError,
  which propagates out of load_payloads() and aborts loading the whole mission.

Concrete current motivating case: the shipping ED F-100D module ships
CoreMods/aircraft/F-100D/UnitPayloads/F-100D.lua. That directory is on the
payload search path, so the file globs -- but its content isn't matched by the
["unitType"] regex, so it isn't cached. The result: loading any mission that
references an F-100D crashes in load_payloads().

## Relationship to #317

Issue #317 ("replace our lua parser with lupa?") diagnoses the same underlying
cause -- the regex-based payload scanner can't handle certain payload-file
syntaxes (the F-100D among them) -- and proposes a larger lupa-based parser
rewrite.

This PR is the small, complementary, defensive fix: it does not attempt to fix
the parser or make the uncacheable file parse. It only stops one
unparseable-but-globbed file from taking down the entire mission load, regardless
of whether or when the larger parser rewrite in #317 lands. A file the scanner
can't cache is simply skipped (which is already the intended outcome for a file
that doesn't match a unit) instead of raising.

## Fix

One call, in load_payloads():

```diff
-                if FlyingType._payload_cache[payload_path] == cls.id and payload_path.exists():
+                if FlyingType._payload_cache.get(payload_path) == cls.id and payload_path.exists():
```

An uncached path -> .get() returns None -> None != cls.id -> the file is
skipped. Cached files are unaffected; nothing else changes.

## Test

Added test_load_payloads_skips_globbed_but_uncached_file to
tests/test_unittype.py (matching the existing tmp_path/monkeypatch style).
It creates a payload directory containing:

- an uncacheable .lua (no ["unitType"]="..." line -- the F-100D failure
  shape), and
- a normal cacheable payload .lua for the unit under test,

points the scanner at the temp directory, and asserts that load_payloads() does
not raise and still loads the valid payload. The test fails with KeyError
against the pre-fix bare subscript and passes with .get().

### DIFF
--- a/dcs/unittype.py
+++ b/dcs/unittype.py
@@ -137,7 +137,7 @@ class FlyingType(UnitType):
             if not payload_dir.exists():
                 continue
             for payload_path in payload_dir.glob("*.lua"):
-                if FlyingType._payload_cache[payload_path] == cls.id and payload_path.exists():
+                if FlyingType._payload_cache.get(payload_path) == cls.id and payload_path.exists():
                     try:
                         payload_main = lua.loads(payload_path.read_text(), _globals=FlyingType._UnitPayloadGlobals)
                     except SyntaxError:
--- a/tests/test_unittype.py
+++ b/tests/test_unittype.py
@@ -1,11 +1,15 @@
 import textwrap
 from pathlib import Path
 
+import pytest
+
 import dcs.countries
 from dcs.liveries.livery import Livery
 from dcs.liveries.liverycache import LiveryCache
 from dcs.liveries.liveryscanner import LiveryScanner
+from dcs.payloads import PayloadDirectories
 from dcs.planes import F_16C_50
+from dcs.unittype import FlyingType
 
 
 def test_plane_liveries(tmp_path: Path) -> None:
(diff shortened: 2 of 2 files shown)


rp-:
https://github.com/pydcs/dcs/pull/439

was faster, sorry

